PR #1293, opened 2023-05-10
[DEV-1692] Allow log level to be configured using environment variable

YS-L:
## Description

This allows logger's log level to be configured using the environment variable `LOG_LEVEL`. Previously, the log level was set based on the configurations yaml file, but that file is no longer mounted into the containers as it should contain only client side configurations.

## Type of Change

## Checklist


docker/entrypoint.sh
```diff
@@ -49,7 +49,7 @@ _main() {
       celery --app featurebyte.worker.start.celery worker -Q io_task,io_task:1,io_task:2,io_task:3 --loglevel=INFO --pool=gevent -c 1000
     elif [ "$1" = 'server' ]; then
       python /scripts/migration.py
-      uvicorn featurebyte.app:app --host=$API_HOST --port=$API_PORT --timeout-keep-alive=300 --log-level=$LOG_LEVEL
+      uvicorn featurebyte.app:app --host=$API_HOST --port=$API_PORT --timeout-keep-alive=300 --log-level=info
```
kchua78, line 52: Do we want to fix the log level for uvicorn?

YS-L: Is it useful to be able to override the log level for uvicorn? If so, I think we can add a different variable to control that. Same for workers' log level which is currently also fixed as INFO.

kchua78: I guess not. Let's leave it as info then.

YS-L:
Timing seems to be back to normal now. Will merge.

PR #1295, opened 2023-05-10
[DEV-1683] Add special handling to query graph nodes with side effect

gee-senbong:
## Description

This PR aims to fix the feature definition bug so that the graph generated by the feature definition is consistent with the query graph used to generated the feature definition.

Currently, query graph node like `AssignNode`, `AliasNode` and `ConditionalNode` modify existing variable in place. This PR fixes this by checking whether the input to these nodes are used by other node. If there exists other node using the same input, it will make a copy to the input node before updating the variable.

## Type of Change

## Checklist


tests/fixtures/sdk_code/feature_with_cleaned_tz_offset_column.py
```diff
@@ -20,8 +20,9 @@
 )
 col = event_view["event_timestamp"]
 col_1 = event_view["tz_offset"]
-event_view["timestamp_hour"] = col.dt.tz_offset(col_1).hour
-grouped = event_view.groupby(
+view = event_view.copy()
```
gee-senbong, line 23: Event view graph node is used by 
* project nodes (to project `event_timestamp` and `tz_offset`)
* assign node

Since it is used by other nodes, a copy is created before updating the frame.

tests/fixtures/sdk_code/feature_operating_system.py
```diff
@@ -0,0 +1,20 @@
...
+scd_view["os_type"] = "unknown"
+col = scd_view["os_type"]
+scd_view["os_type"][col.str.contains(pat="window", case=True)] = "window"
+col_1 = scd_view["os_type"]
+scd_view["os_type"][col.str.contains(pat="mac", case=True)] = "mac"
+grouped = scd_view.as_features(
+    column_names=["os_type"], feature_names=["os_type"], offset=None
+)
+feat = grouped["os_type"]
+output = feat
```
gee-senbong, line 20: For the statement like
```
col_1 = scd_view["os_type"]
```
We could use `vulture` package (https://github.com/jendrikseipp/vulture) to detect & remove unused statements.

gee-senbong: Seems like internally, `vulture` is using `ast`. Will update the SDK code generation logic in the next PR to remove unused statements.

tests/unit/api/test_sdk_code.py
```diff
@@ -337,3 +337,80 @@ def test_sdk_code_generation__item_view_cosine_similarity_feature(
...
+        item_table_id=saved_item_table.id,
+    )
+
+
+def test_sdk_code_generation__operating_system_feature(
+    saved_scd_table, cust_id_entity, update_fixtures
+):
+    """
+    Test get_feature_jobs_status for feature without tile
+    """
```
gee-senbong, line 399: Will update the doc string here.

featurebyte/query_graph/node/generic.py
```diff
@@ -1418,22 +1463,55 @@ def derive_var_type(self, inputs: List[OperationStructure]) -> DBVarType:
...
+        if context.as_info_str:
+            # This is to handle the case where `View[<column>][<condition>] = <value>` is used.
+            # Since there is no single line in SDK code to generate conditional expression, we output info instead
+            # and delay the generation of SDK code to the assign node. This method only generates the conditional part,
+            # the assignment part will be generated in the assign node.
+            return statements, InfoStr(
+                json.dumps({"value": self.parameters.value, "mask": mask_var_name})
+            )
+
+        # This handles the normal series assignment case.
```
YS-L, line 1515: Maybe this comment can clarify what normal series assignment means. Does it mean series assignment that doesn't affect parent view?

featurebyte/query_graph/node/metadata/sdk_code.py
```diff
@@ -74,6 +74,12 @@ def as_input(self) -> str:
         return str(self)
 
 
+class InfoStr(str):
+    """
+    InfoStr class is used to represent a string that contains information about the code.
```
YS-L, line 79: Does `InfoStr` has to be a string? It seems responsible for passing on some key value information for subsequent node's code generation, rather than being converted to code directly by itself. What do you think about having this class handle the to and from json? Is that conversion to json still required if this is not a str?

featurebyte/query_graph/enum.py
```diff
@@ -96,6 +96,18 @@ class NodeType(StrEnum):
...
+    def inplace_operation_node_type(cls) -> Set["NodeType"]:
+        """
+        Returns all the node types that are inplace operations. Inplace operations are operations that
+        modify the input node inplace.
+
+        Returns
+        -------
+        Set[NodeType]
+        """
+        return {cls.ASSIGN, cls.CONDITIONAL, cls.ALIAS}
```
YS-L, line 109: What do you think of making this a node level attribute / property that each node has to define?

CHANGELOG.md
```diff
@@ -30,6 +30,7 @@
 ### 🧰 Bug fixes 🧰
 
 * Removed additional dependencies specified in featurebyte client
+* Fixed bug in `feature.definition` so that it is consistent with the underlying query graph
```
YS-L, line 33: Since 0.2.2 is already out, I think this should be moved up to the unreleased version section.

featurebyte/query_graph/transform/sdk_code.py
```diff
@@ -40,11 +39,103 @@ class SDKCodeGlobalState(BaseModel):
...
+        backward_nodes: List[Node]
+            Backward nodes of the node
+        edges_map: Dict[str, str]
+            Edges map of the query graph
+        """
+        if (
+            node.type in NodeType.inplace_operation_node_type()
+            and len(set(edges_map[backward_nodes[0].name])) > 1
+        ):
+            # If there are more than one distinct output nodes from the first input, it means that the backward node
```
YS-L, line 114: I think including an example here would be helpful.

featurebyte/query_graph/enum.py
gee-senbong: I think that is even better as we force developer to think about this when inheriting the `BaseNode` class. Will move this to the node class.

featurebyte/query_graph/node/metadata/sdk_code.py
gee-senbong: This is a good point. The initial reason to inherit as a string is for the experimentation of the fix. Will change the type to something like `InfoDict` which inherit dictionary. Doing do, we can avoid unnecessary conversion between string & dict.

CHANGELOG.md
gee-senbong: Good catch! Will fix this.

featurebyte/query_graph/transform/sdk_code.py
```diff
@@ -40,11 +39,137 @@ class SDKCodeGlobalState(BaseModel):
...
+            assign_input_view_op_struct = node_name_to_operation_structure[
+                assign_node_input_node_names[0]
+            ]
+            assign_input_view_columns = assign_input_view_op_struct.output_column_names
+            assign_column_name = assign_node.parameters.name  # type: ignore
+            if assign_column_name in assign_input_view_columns:
+                # This handle the case `view[<col>][<mask>] = <value>`. Since this SDK code actually contains
+                # ConditionalNode and AssignNode, we need to output info string for the ConditionalNode and
+                # postpone the assignment operation to AssignNode.
+                self.as_info_dict_node_names.add(node.name)
```
gee-senbong, line 98: @YS-L I identify another case where previous fix does not consider. Update the logic here so that it can differentiate these two cases.

tests/unit/api/test_sdk_code.py
```diff
@@ -337,3 +338,117 @@ def test_sdk_code_generation__item_view_cosine_similarity_feature(
...
+    col[mask] = "new_value"
+    assert view != view_copy  # view becomes `[input] -> [graph] -> [assign]`
+
+    # check that no way we could get the handle to conditional node from the api object.
+    # if we could get the handle and user does something like `view_copy["col_text"] = <handle>`.
+    # this will break the assumption used in the sdk code generation to identify case 1 (view-mask-assign)
+    # and case 2 (view-series-assignment) conditional assignment.
+    assert view.node_name != view_copy.node_name
+    assert view_copy["col_text"].node.type == NodeType.PROJECT  # before conditional node
+    assert view["col_text"].node.type == NodeType.PROJECT  # after assignment node
```
gee-senbong, line 454: @YS-L I also updated `test_sdk_code_generation__operating_system_feature` test and add a test to cover the assumption here.

gee-senbong: Unless user does something like
```
col = view["col_text"]
col._parent = None
col[mask] = "new_value"
assert col.node.type == Node.Conditional
```

featurebyte/query_graph/transform/sdk_code.py
```diff
@@ -40,11 +39,137 @@ class SDKCodeGlobalState(BaseModel):
...
+            Backward nodes of the node
+        edges_map: Dict[str, str]
+            Edges map of the query graph
+        """
+        if (
+            node.is_inplace_operation_in_sdk_code
+            and len(set(edges_map[backward_nodes[0].name])) > 1
+        ):
+            # If there are more than one distinct output nodes from the first input, it means that the backward node
+            # is used by other nodes. In this case, we need to copy the backward node to avoid inplace operation
```
YS-L, line 141: By `copy the backward node`, does it actually mean copying the node or the view? I think it's better to update the comment since it gets a bit confusing.

featurebyte/query_graph/transform/sdk_code.py
```diff
@@ -40,11 +39,137 @@ class SDKCodeGlobalState(BaseModel):
...
+            # affecting other nodes. Currently, only first node in the backward node's forward nodes will be
+            # affected by the inplace operation. For example, if we have an AssignNode that consume input `view`
+            # it will generate `view[<col>] = <value>` SDK code if the `view` object is not used by other nodes.
```
YS-L, line 144: What does "`view` object is not used by other nodes" mean? It would be great if the example can show two cases of SDK code snippets where one requires copy, and another doesn't require copy.

featurebyte/query_graph/transform/sdk_code.py
gee-senbong: It should be the input `view` but not `backward node`. Will update this part.

featurebyte/query_graph/transform/sdk_code.py
gee-senbong: Will update this part to include example with more complete SDK code. Is there any other update needed as it takes longer time to run the complete tests. Would like to get this PR merged first as it block other works like removing unused copy statements.

gee-senbong: Updated the comment here with an example.

YS-L (approved):

PR #1300, opened 2023-05-11
[DEV-1699] Avoid performing Spark storage connection check for every session creation

kchua78:
## Description

- Connection check is expensive - do it only when registering objects instead of each time a session is created.
- Fix GCS credentials not encrypted for storage

## Related Issue

https://featurebyte.atlassian.net/browse/DEV-1699

## Type of Change

## Checklist

### DIFF
--- a/featurebyte/models/credential.py
+++ b/featurebyte/models/credential.py
@@ -73,6 +73,15 @@ class BaseCredential(FeatureByteBaseModel):
         for field in self.__fields__.values():
             if field.type_ == StrictStr:
                 setattr(self, field.name, encrypt_value(getattr(self, field.name)))
+            elif field.type_ == str:
+                field_value = getattr(self, field.name)
+                if isinstance(field_value, dict):
+                    # Encrypt each value in the dict
+                    setattr(
+                        self,
+                        field.name,
+                        {key: encrypt_value(value) for key, value in field_value.items()},
+                    )
 
     def decrypt(self) -> None:
         """
@@ -81,6 +90,15 @@ class BaseCredential(FeatureByteBaseModel):
         for field in self.__fields__.values():
             if field.type_ == StrictStr:
                 setattr(self, field.name, decrypt_value(getattr(self, field.name)))
+            elif field.type_ == str:
+                field_value = getattr(self, field.name)
+                if isinstance(field_value, dict):
+                    # Decrypt each value in the dict
+                    setattr(
+                        self,
+                        field.name,
+                        {key: decrypt_value(value) for key, value in field_value.items()},
+                    )
 
 
 # Database Credentials
--- a/featurebyte/models/credential.py
+++ b/featurebyte/models/credential.py
@@ -74,6 +74,7 @@ class BaseCredential(FeatureByteBaseModel):
             if field.type_ == StrictStr:
                 setattr(self, field.name, encrypt_value(getattr(self, field.name)))
             elif field.type_ == str:
+                # pydantic captures dict field type as str
                 field_value = getattr(self, field.name)
                 if isinstance(field_value, dict):
                     # Encrypt each value in the dict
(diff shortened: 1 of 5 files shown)


featurebyte/session/base_spark.py
```diff
@@ -336,9 +343,13 @@ async def list_procedures(self) -> list[str]:
         return []
 
     async def register_missing_objects(self) -> None:
+        # check storage connection is working
+        session = cast(BaseSparkSession, self.session)
+        session.test_storage_connection()
```
YS-L, line 348: After moving this here, storage connection will still be tested on feature store creation, is that right?

featurebyte/models/credential.py
```diff
@@ -73,6 +73,15 @@ def encrypt(self) -> None:
         for field in self.__fields__.values():
             if field.type_ == StrictStr:
                 setattr(self, field.name, encrypt_value(getattr(self, field.name)))
+            elif field.type_ == str:
```
YS-L, line 76: Is this a new behaviour and is it covered in tests?

featurebyte/models/credential.py
```diff
@@ -73,6 +73,15 @@ def encrypt(self) -> None:
...
+                field_value = getattr(self, field.name)
+                if isinstance(field_value, dict):
```
YS-L, line 78: Can this still evaluate to true given the `field.type_ == str` condition above?

featurebyte/session/base_spark.py
kchua78: Yes it will be run every time we re register the udfs

featurebyte/models/credential.py
kchua78: This is covered by the credential model unit tests for encryption functionality. Without this change the tests that use GCS storage credential will fail

featurebyte/models/credential.py
kchua78: It's strange for me too seems the pydantic `field.type_` attribute is str even though the field is declared as `Dict[str, str]`

featurebyte/models/credential.py
YS-L: Is it possible to have a unit test for this?

featurebyte/models/credential.py
YS-L: I see. Maybe worth adding a comment. Btw curious why the GCSStorageCredential structure has a nested dictionary structure, as opposed to a flattened one like S3StorageCredential?

YS-L (approved):

featurebyte/models/credential.py
kchua78: There is already a test. It was not covered previously because GCS credential is not in the test parameters but that is updated in this PR

tests/unit/models/test_credential.py
```diff
@@ -12,14 +12,17 @@
...
     S3StorageCredential,
     StorageCredentialType,
     UsernamePasswordCredential,
 )
 from featurebyte.models.feature_store import FeatureStoreModel
 
 
-@pytest.fixture(name="storage_credential", params=[None, StorageCredentialType.S3])
+@pytest.fixture(
+    name="storage_credential", params=[None] + list(StorageCredentialType.__members__.values())
```
kchua78, line 24: This change will make sure all storage credentials are covered in the tests

featurebyte/models/credential.py
kchua78: The GCS service account is a json dict (not nested). It's usually provided as a file path but since we need to store this in the db that won't work. Another option is to change the parameter into a json string. The client that we use accepts either a file path or a python dictionary

featurebyte/models/credential.py
YS-L: Sorry, I missed that!

PR #1324, opened 2023-05-17
[DEV-1719] Call F_TIMESTAMP_TO_INDEX directly in tile generation sql

YS-L:
## Description

This refactors the tile generation sql to call `F_TIMESTAMP_TO_INDEX` directly to generate the final tile indices. This removes an assumption that there is a fixed tile start date in the tile sql, and the resulting sql is actually simpler. 

This will enable an optimisation in tile cache to reduce unnecessary recalculation of tiles (to be added in a next PR). In that optimisation, different entities can have different tile start dates.

## Type of Change

## Checklist


tests/unit/query_graph/test_interpreter.py
```diff
@@ -312,20 +312,14 @@ def test_graph_interpreter_on_demand_tile_gen(
     expected_sql = textwrap.dedent(
         f"""
         SELECT
-          TO_TIMESTAMP(
-            DATE_PART(EPOCH_SECOND, CAST(__FB_START_DATE AS TIMESTAMPNTZ)) + tile_index * 3600
-          ) AS __FB_TILE_START_DATE_COLUMN,
```
YS-L, line 317: Previously, `F_TIMESTAMP_TO_INDEX` is called on this intermediate column later by tile manager. But the function works the same way if event timestamp column is passed in directly, so creating this intermediate column is not necessary.

featurebyte/query_graph/sql/tile_compute.py
```diff
@@ -80,22 +79,14 @@ def construct_tile_sqls(self) -> dict[str, Select]:
...
+            tile_sql_expr = get_tile_sql_from_point_in_time(
                 sql_template=tile_info.sql_template,
                 point_in_time_list=self.point_in_time_list,
                 frequency=tile_info.frequency,
                 time_modulo_frequency=tile_info.time_modulo_frequency,
                 blind_spot=tile_info.blind_spot,
                 window=self.get_max_window_size(tile_info.tile_table_id),
             )
-            # Include global tile index that would have been computed by F_TIMESTAMP_TO_INDEX UDF
-            # during scheduled tile jobs
```
YS-L, line 92: The query above already contains the global tile index, so the following is no longer needed.

featurebyte/query_graph/sql/ast/tile.py
```diff
@@ -36,7 +36,10 @@ class BuildTileNode(TableNode):
     value_by: str | None
     tile_specs: list[TileSpec]
     timestamp: str
-    frequency: int
+    frequency_minute: int
+    blind_spot: int
+    time_modulo_frequency: int
```
gee-senbong, line 41: What does `frequency_minute` mean? Would be good to leave some comments on the input to `F_TIMESTAMP_TO_INDEX` here.

featurebyte/query_graph/sql/ast/tile.py
```diff
@@ -46,29 +49,30 @@ class BuildTileNode(TableNode):
...
+                expressions=[
+                    self.context.adapter.convert_to_utc_timestamp(
+                        quoted_identifier(self.timestamp)
+                    ),
+                    make_literal_value(self.time_modulo_frequency),
+                    make_literal_value(self.blind_spot),
+                    make_literal_value(self.frequency_minute),
+                ],
+            ),
+            alias="index",
```
gee-senbong, line 65: Do we need to worry whether `index` may cause name collision here?

gee-senbong (approved):
Left some minor comments to consider. In general, LGTM.

yanhuisg (approved):
LGTM

featurebyte/query_graph/sql/ast/tile.py
YS-L: This is the frequency component of the feature job settings but using minute as the unit instead of second. This is a requirement of the `F_TIMESTAMP_TO_INDEX` function. Will add comments in a next PR.

featurebyte/query_graph/sql/ast/tile.py
YS-L: Yes, that is an existing issue. It should be replaced with an internal column name: https://featurebyte.atlassian.net/browse/DEV-1730

PR #1333, opened 2023-05-19
[DEV-1721] Enforce time limits for periodic tasks scheduled based on intervals

kchua78:
## Description

Periodic tasks scheduled based on intervals can pile up if such task does not complete within the interval.

Changes made in this PR:

- Always specify soft time limit for scheduled tasks (defaults to interval length), can be overriden
- Support enforcement of soft time limit for io-intensive tasks executed in gevent workers (soft limit is not supported by celery for gevent worker pool)

## Related Issue

https://featurebyte.atlassian.net/browse/DEV-1721

## Type of Change

## Checklist

- [x] I have read the [`CODE_OF_CONDUCT.md`](https://github.com/featurebyte/featurebyte/blob/main/CODE_OF_CONDUCT.md) and [`CONTRIBUTING.md`](https://github.com/featurebyte/featurebyte/blob/main/CONTRIBUTING.md) guides.
- [x] I have written tests for the changes made.
- [x] I have written docstrings in [NumpyDoc format](https://numpydoc.readthedocs.io/en/latest/format.html#docstring-standard)
- [x] I have labeled my Pull Request correctly

### DIFF
--- a/featurebyte/worker/task_executor.py
+++ b/featurebyte/worker/task_executor.py
@@ -60,7 +60,7 @@ def run_async(coro: Awaitable[Any], timeout: Optional[int] = None) -> Any:
         timeout is exceeded
     """
     try:
-        loop = asyncio.get_event_loop()
+        loop = asyncio.get_running_loop()
         logger.debug("Use existing async loop", extra={"loop": loop})
     except RuntimeError:
         loop = asyncio.new_event_loop()
--- a/featurebyte/worker/task_executor.py
+++ b/featurebyte/worker/task_executor.py
@@ -3,7 +3,7 @@ This module contains TaskExecutor class
 """
 from __future__ import annotations
 
-from typing import Any, Awaitable
+from typing import Any, Awaitable, Optional
 
 import asyncio
 from concurrent.futures import ThreadPoolExecutor
@@ -11,6 +11,7 @@ from threading import Thread
 from uuid import UUID
 
 import gevent
+from celery.exceptions import SoftTimeLimitExceeded
 
 from featurebyte.enum import WorkerCommand
 from featurebyte.logging import get_logger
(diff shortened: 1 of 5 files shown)


featurebyte/worker/task_executor.py
```diff
@@ -61,13 +69,20 @@ def run_async(coro: Awaitable[Any]) -> Any:
...
+    try:
+        with gevent.Timeout(seconds=timeout, exception=TimeoutError):
+            event = gevent.event.Event()
+            future.add_done_callback(lambda _: event.set())
+            event.wait()
+            return future.result()
+    except TimeoutError as exc:
+        # try to cancel the job if it has not started
+        future.cancel()
+        raise SoftTimeLimitExceeded(f"Task timed out after {timeout}s") from exc
```
gee-senbong, line 85: Will it throw another exception (when calling `future.cancel()`) if the job is started already?

featurebyte/worker/task_executor.py
```diff
@@ -61,13 +69,20 @@ def run_async(coro: Awaitable[Any]) -> Any:
...
-    event = gevent.event.Event()
-    future.add_done_callback(lambda _: event.set())
-    event.wait()
-    return future.result()
+    try:
+        with gevent.Timeout(seconds=timeout, exception=TimeoutError):
+            event = gevent.event.Event()
+            future.add_done_callback(lambda _: event.set())
+            event.wait()
+            return future.result()
```
gee-senbong, line 81: Seems like we could pass the `timeout` parameter to `future.result()` here (https://docs.python.org/3/library/asyncio-task.html#scheduling-from-other-threads).

featurebyte/worker/task_executor.py
kchua78: cancel will not raise any exception but return False if the task cannot be cancelled

featurebyte/worker/task_executor.py
kchua78: Tried that - it raises timeout error but unfortunately the async task continues to run in the asyncio thread

gee-senbong (approved):

PR #1362, opened 2023-05-25
[chore] Fix bug in payload fixture generations

gee-senbong:
## Description

As per title, this PR aims to fix bug in request payload test fixtures generation.

## Type of Change

## Checklist


kchua78 (approved):

Taskfile.yaml
```diff
@@ -201,7 +201,7 @@ tasks:
       - featurebyte/**/*
       - tests/**/*
     cmds:
-      - poetry run pytest --timeout=240 --junitxml=pytest.xml.0 -n auto --cov=featurebyte tests/unit
+      - poetry run pytest --timeout=240 --junitxml=pytest.xml.0 -n auto --cov=featurebyte tests/unit tests/unit/conftest.py
```
bushwhackr, line 205: just wondering why you have to specify the conftest directly. Isn't the previous `tests/unit` already specified the conftest?

gee-senbong: If we don't specify the conftest, the test won't run. That's the reason why the issue is not caught. Another option is to move the test to generate the fixtures into a file starts with `test_*`.

YS-L (approved):

bushwhackr (approved):

Taskfile.yaml
```diff
@@ -202,6 +202,7 @@ tasks:
       - tests/**/*
     cmds:
       - poetry run pytest --timeout=240 --junitxml=pytest.xml.0 -n auto --cov=featurebyte tests/unit
+      - poetry run pytest --timeout=240 --update-fixtures tests/unit/test_generate_payload_fixtures.py
```
YS-L, line 205: Would this regenerate the fixtures every time when running `task test-unit`? Is that the desired behaviour? I thought the `--update-fixtures` command is something that should be run manually only when needed.

gee-senbong: Yes. The main idea here is to run the actual fixture generation code (it will generate the fixtures but the changes will not be committed). Does it make sense to you?

YS-L: It seems a bit inconvenient for developers that the command results in many changed files that should not be committed when running locally. How about running without the `--update-fixtures` flag? Then, it would still run the fixture generation test but without writing out the fixtures. Would that catch the issue fixed by this PR?

bushwhackr (approved):
i'll update the infrastrucutre repo

YS-L (approved):

kchua78 (approved):

.github/workflows/test_gen.yaml
```diff
@@ -109,7 +109,9 @@ jobs:
         name: hive-udf
         path: featurebyte/sql/spark/
     - name: Run tests
-      run: task test-unit
+      run: |-
+        task test-unit
+        task generate-unit-test-fixtures
```
gee-senbong, line 114: Run the `generate-unit-test-fixtures` after `test-unit` so that the newly generated fixtures won't affect the existing unit tests.

PR #1373, opened 2023-05-30
[DEV-1756] Improve columns batching logic to account for aggregation parameters

YS-L:
## Description

This improves the columns batching logic in historical features to account for aggregation parameters so that it is not sensitive to the original ordering of features. The change is to introduce a reordering step in the `split_nodes` function.

The previously measured historical features runtime for [`Features V1`](https://github.com/featurebyte/featurebyte-notebooks/blob/f6e8cd4b352c3550540874e9cdd9001ba47a84f0/notebooks/features/grocery/setup.py#L67) was optimistic because the features were ordered meaningfully in the feature list. The runtime increases from 27m to 42m if the feature list is shuffled randomly. With the changes in this PR, the runtime of the shuffled feature list is 25m.

## Type of Change

## Checklist

### DIFF
--- a/featurebyte/query_graph/sql/feature_historical.py
+++ b/featurebyte/query_graph/sql/feature_historical.py
@@ -35,6 +35,7 @@ from featurebyte.query_graph.sql.common import (
 )
 from featurebyte.query_graph.sql.feature_compute import FeatureExecutionPlanner
 from featurebyte.query_graph.sql.parent_serving import construct_request_table_with_parent_entities
+from featurebyte.query_graph.sql.specs import NonTileBasedAggregationSpec, TileBasedAggregationSpec
 from featurebyte.session.base import BaseSession
 from featurebyte.tile.tile_cache import TileCache
 
@@ -365,8 +366,10 @@ def get_historical_features_expr(
 
 
 def split_nodes(
+    graph: QueryGraph,
     nodes: list[Node],
     num_features_per_query: int,
+    is_tile_cache: bool = False,
 ) -> list[list[Node]]:
     """
     Split nodes into multiple lists, each containing at most `num_features_per_query` nodes. Nodes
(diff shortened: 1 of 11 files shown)


tests/fixtures/expected_historical_requests_multiple_batches_feature_set_0.sql
```diff
@@ -134,14 +32,44 @@ WITH "REQUEST_TABLE_W7200_F3600_BS900_M1800_CUSTOMER_ID" AS (
...
 )
 SELECT
   AGG."__FB_ROW_INDEX_FOR_JOIN",
   AGG."POINT_IN_TIME",
   AGG."CUSTOMER_ID",
-  "_fb_internal_window_w7200_avg_30d0e03bfdc9aa70e3001f8c32a5f82e6f793cbb" AS "a_2h_average",
-  "_fb_internal_window_w172800_avg_30d0e03bfdc9aa70e3001f8c32a5f82e6f793cbb" AS "a_48h_average",
+  "_fb_internal_as_at_count_None_membership_status_None_input_2" AS "asat_feature",
   "_fb_internal_item_count_None_order_id_None_input_1" AS "order_size"
 FROM _FB_AGGREGATED AS AGG
```
gee-senbong, line 75: Seems like the SQL becomes shorter & more readable than before.

featurebyte/query_graph/sql/feature_historical.py
```diff
@@ -365,33 +366,68 @@ def get_historical_features_expr(
...
+        key = ",".join(parts)
+        return key
+
     result = []
-    for i in range(0, len(nodes), num_features_per_query):
-        current_nodes = nodes[i : i + num_features_per_query]
+    sorted_nodes = sorted(nodes, key=get_sort_key)
+    for i in range(0, len(sorted_nodes), num_features_per_query):
+        current_nodes = sorted_nodes[i : i + num_features_per_query]
         result.append(current_nodes)
```
gee-senbong, line 424: Would you expect to see performance improvement if we could ensure that same source table should be within the same node group (if the size of the group is less than `num_features_per_query`)?

gee-senbong (approved):
LGTM. Should we add a change log entry for this PR?

featurebyte/query_graph/sql/feature_historical.py
YS-L: If the size of the group is less than `num_features_per_query`, all the nodes would have been in the same group already regardless of source table, so I think I'm misunderstanding the question. Could you clarify a bit more?

YS-L:
Regarding the changelog, I think the changes here are effectively part of this entry: https://github.com/featurebyte/featurebyte/blob/b908604e4d2a8858c5e1b63b69a3bbddfa191a5d/.changelog/DEV-1740.yaml#L19

since the above change is not published in any version yet.

PR #1378, opened 2023-05-31
[DEV-1705] Kerberos auth and webhdfs support for spark session

kchua78:
## Description

Add kerberos authentication and webhdfs support for spark session.

## Related Issue

https://featurebyte.atlassian.net/browse/DEV-1705

## Type of Change

## Checklist


bushwhackr (approved):
LGTM

featurebyte/session/spark.py
```diff
@@ -30,46 +34,75 @@ class SparkSession(BaseSparkSession):
...
+        # so that we can use kerberos cache to authenticate storage
+        database_credential = data.get("database_credential")
+        if database_credential:
+            if isinstance(database_credential, KerberosKeytabCredential):
+                # create a temporary keytab file to specify the KDC
+                with tempfile.NamedTemporaryFile(mode="wb", suffix=".keytab") as keytab_file:
+                    keytab_file.write(database_credential.keytab)
+                    keytab_file.flush()
+                    cmd = ["kinit", "-kt", keytab_file.name, database_credential.principal]
+                    subprocess.run(cmd, check=True)
```
YS-L, line 61: Will there be any concerns launching a subprocess if this code runs in the API server / an async io task?

tests/integration/session/test_simple_storage.py
```diff
@@ -50,3 +61,35 @@ def test_azure_blob_storage():
         ),
     )
     storage.test_connection()
+
+
+@pytest.mark.skip(reason="Kerberos is not supported in CI")
```
YS-L, line 66: Is there a plan to support testing this in an automated way?

featurebyte/session/spark.py
kchua78: it should not be an issue (we run subprocess in async pytests as well) but we definitely should test this in a deployed service to be sure

kchua78: Tested end-to-end (i.e. register store, initialize schema and preview source tables). 
Found some bugs and fixed them.

tests/integration/session/test_simple_storage.py
kchua78: Once possibility would be to maintain a single-node kerberized cluster for integration testing. The other would be to create a container for testing but it may be rather large to pull for PR tests.

featurebyte/session/spark.py
```diff
@@ -40,47 +50,56 @@ class SparkSession(BaseSparkSession):
...
+                database=self.schema_name,
+                port=self.port,
+                access_token=access_token,
+                auth=auth,
+                scheme=scheme,
+                kerberos_service_name=kerberos_service_name,
+            )
+
+        # create a connection
+        try:
```
kchua78, line 93: Refactor this to avoid running kinit on each session creation. Instead only run kinit on failure (e.g. ticket does not exist or expired)

featurebyte/routes/feature_store/controller.py
```diff
@@ -96,52 +100,51 @@ async def create_feature_store(
...
+            await self.session_validator_service.validate_feature_store_id_not_used_in_warehouse(
+                feature_store_name=data.name,
+                session_type=data.type,
+                details=data.details,
+                get_credential=get_credential_to_use,
+                users_feature_store_id=document.id,
+            )
+            logger.debug("End validate_feature_store_id_not_used_in_warehouse")
 
-        async with self.service.persistent.start_transaction():
```
kchua78, line 143: @jevonyeoh Replaced transaction with try / catch + clean up here as a transaction times out in 60s. For slow connections this can sometimes be an issue.

jevonyeoh: sounds good - any reason not to place the `try` line here instead where the txn previously was? that way the lines above won't have to be changed, and the scope of the `try-catch` block is as narrow as possible

featurebyte/session/spark.py
```diff
@@ -4,72 +4,148 @@
...
-        super().__init__(**data)
-
         auth = None
         scheme = None
+        access_token = None
+        kerberos_service_name = None
+
+        super().__init__(**data)
+
+        if self.database_credential:
```
jevonyeoh, line 63: would it be cleaner to have separate `SparkSession#fromToken` and `SparkSession#fromKerberos` methods, instead of trying to overload the constructor this way?

kchua78: It makes sense if we are instantiating these classes manually. However since these classes are instantiated by the session manager, it will make the logic there more complex.

featurebyte/session/webhdfs.py
```diff
@@ -0,0 +1,230 @@
...
+            raise WebHdfsException(msg=self._response.text, status_code=self._response.status_code)
+        self._buf = b""
+
+
+class BufferedOutput(BufferedOutputBase):  # pylint: disable=too-many-instance-attributes
+    """
+    BufferedOutput class for webhdfs with Kerberos authentication
+    """
+
+    def __init__(  # pylint: disable=super-init-not-called
```
jevonyeoh, line 161: curious why we don't call the the parent init?

jevonyeoh (approved):

PR #1389, opened 2023-06-02
[DEV-1777] Refactor task manager to take celery object as input

kchua78:
## Description

Refactor task manager to make celery an input similar to persistent, storage etc.

## Related Issue

https://featurebyte.atlassian.net/browse/DEV-1777

## Type of Change

## Checklist


featurebyte/worker/__init__.py
```diff
@@ -8,47 +8,76 @@
...
+    database_name: str
+        Database name
+
+    Returns
+    -------
+    Celery
+    """
+    global CELERY
+    if CELERY:
+        return CELERY
```
gee-senbong, line 35: Do we need to consider whether this is thread-safe or not?

gee-senbong: Noted that it is not related to the current refactoring PR. Just check whether we need to fix this. If the construction of the celery object is light, would prefer to make it like the `get_persistent` and don't store the global object. It is also more future proof in case someone in the future introduce some states to the celery instance.

kchua78: There is some cost to calling the function each time, but it's not very large
```
>>> timeit.timeit("get_celery()", globals=globals(), number=1000)
0.046812791999997216
>>> timeit.timeit("celery", globals=globals(), number=1000)
4.016700000875062e-05
```

featurebyte/app.py
```diff
@@ -87,6 +88,7 @@ def _dep_injection_func(
             task_manager=TaskManager(
                 user=request.state.user,
                 persistent=request.state.persistent,
+                celery=get_celery(),
```
kchua78, line 91: @gee-senbong Getting a new celery instance in each request

pyproject.toml
```diff
@@ -136,6 +136,7 @@ sqlglot = "^10.1.3,<10.4"  # SQL generation doesn't match as >10.4 double quotes
 thrift-sasl = { version = "^0.4.3", optional = true }
 typeguard = "^2.13.3"
 typer = "^0.7.0"
+typing-extensions = "4.5.0"
```
kchua78, line 139: @gee-senbong Newer versions of typeguard does not seem to work in darwin env (seems to work in linux). Pinning this for now so we can investigate separately.

tests/integration/conftest.py
```diff
@@ -1306,7 +1306,7 @@ def mock_task_manager(request, persistent, storage, temp_storage, get_cred, mock
     Mock celery task manager for testing
     """
     _ = mock_app_callbacks
-    if "disable_task_manager_mock" in request.keywords:
+    if request.module.__name__ == "test_task_manager":
```
kchua78, line 1309: Checking `request.keywords` does not work when the fixture has a module scope, so the tests that required this mock to be skipped did not actually have the mock skipped. Checking for module name instead (only tests in the test_task_manager integration test script runs jobs in an actual celery worker process and need this mock to be skipped)

featurebyte/worker/__init__.py
gee-senbong: Seems like it is costing 4ms on average. We can improve this later if it becomes a bottleneck.

tests/integration/conftest.py
gee-senbong: Good to find this. I encountered some similar failures locally previously. 👍

gee-senbong (approved):
LGTM

PR #1391, opened 2023-06-03
[DEV-1779] Support creation of static source table from source table or view

kchua78:
## Description

Support creation of static source table from source table or view.
This PR only implements CRUD of these static source tables. 
Creation of featurebyte tables from static source table, and creation by uploading a local file will be added in the future.

## Related Issue

https://featurebyte.atlassian.net/browse/DEV-1779

## Type of Change

## Checklist


featurebyte/service/info.py
```diff
@@ -1123,3 +1131,36 @@ async def get_batch_feature_table_info(
             created_at=batch_feature_table.created_at,
             updated_at=batch_feature_table.updated_at,
         )
+
+    async def get_static_source_table_info(
```
jevonyeoh, line 1135: still wondering - is it necessary to have this centralized info service, vs. having each API object's service implement it's own info method?

tests/fixtures/request_payloads/batch_feature_table.json
```diff
@@ -4,9 +4,9 @@
         "Instead, update the test tests/unit/test_generate_payload_fixtures.py#test_save_payload_fixtures.",
         "Run `pytest --update-fixtures` to update it."
     ],
-    "_id": "646f6c1c0ed28a5271fb02da",
-    "batch_request_table_id": "646f6c1c0ed28a5271fb02d9",
-    "deployment_id": "646f6c1c0ed28a5271fb02d6",
-    "feature_store_id": "646f6c190ed28a5271fb02a1",
+    "_id": "647b5ba9875a4313db21a1df",
+    "batch_request_table_id": "647b5ba9875a4313db21a1de",
```
jevonyeoh, line 8: nit: would be nice to be able to "lock" these ID's so that they aren't always regenerated whenever we update fixures

tests/unit/service/test_static_source_table.py
```diff
@@ -0,0 +1,129 @@
...
+    Fixture for a TableDetails
+    """
+    return TableDetails(
+        database_name="fb_database",
+        schema_name="fb_schema",
+        table_name="fb_table",
+    )
+
+
+@pytest.fixture(name="db_session")
```
jevonyeoh, line 60: can we reuse the fixture in `test_observation_table` here?

tests/unit/service/test_static_source_table.py
```diff
@@ -0,0 +1,129 @@
...
+            {"name": "a", "dtype": "INT"},
+            {"name": "b", "dtype": "INT"},
+            {"name": "c", "dtype": "INT"},
+        ],
+        num_rows=1000,
+        most_recent_point_in_time="2023-01-15T10:00:00",
+    )
+
+
+@pytest.fixture(name="table_details")
```
jevonyeoh, line 48: can we reuse the fixture in `test_observation_table` for this too?

tests/fixtures/request_payloads/batch_feature_table.json
kchua78: Indeed. It makes PRs look intimidating

tests/unit/service/test_static_source_table.py
kchua78: Sure. Good idea

tests/fixtures/request_payloads/batch_feature_table.json
kchua78: Will revert these fixture changes for now and manually update the new payload fixture. We can refactor in separate PR since this one is rather large atm

featurebyte/service/info.py
kchua78: I think it was centralized to make it easy to inject info from one asset into that for another asset (e.g. entity info in table info etc. We can discuss and refactor if necessary in a separate effort.

tests/unit/service/test_static_source_table.py
kchua78: I realized that some tests are unnecessary since the service shares the same based class with several other classes that already provided test coverage. Will be removing these tests, and as a result these fixtures are no longer required.

featurebyte/service/info.py
jevonyeoh: i see - in that situation, i think it would still be better if the table service were the only one importing those dependencies, as that would make it clearer as to which objects actually depend on others.

but yea, no action needed for now!

jevonyeoh (approved):
can add to docs, and also potentially expose the `StaticSourceTable` object in the `__init__` file when ready

PR #1425, opened 2023-06-14
enhancement: optimize batch feature payload preparation

gee-senbong:
## Description

Current batch feature payload in featurel list's `_pre_save_operations` method takes long time to prepare as
* it will prune the graph for each feature first
* combined all the pruned graph into a single graph.

To optimize the time taken to save a batch of features, this PR introduces a new `quick_prune` method that consume multiple target nodes & pruning all the targeted features in one go.

## Type of Change

## Checklist


featurebyte/api/feature_list.py
```diff
@@ -677,13 +678,29 @@ def _pre_save_operations(self, conflict_resolution: ConflictResolution = "raise"
                     feat = Feature.get(name=feat_name)
                     self.feature_objects[feat_name] = feat
                 except RecordRetrievalException:
-                    feature_payloads.append(FeatureCreate(**feat.dict(by_alias=True)))
+                    features.append(feat)
             if conflict_resolution == "raise" and not feat.saved:
-                feature_payloads.append(FeatureCreate(**feat.dict(by_alias=True)))
+                features.append(feat)
```
gee-senbong, line 683: **main**
`feat.dict(by_alias)` is expensive as it will prune the feature graph from the global query graph. Each feature pruning will involve one operation structure extraction & reconstruction of the pruned feature graph. `BatchFeatureCreate.create` combines all individual pruned graphs into a single combined graph.

**current branch**
`GlobalQueryGraph().crop(target_node_names=[feat.node_name for feat in features])` directly crops the global graph based on the given list of target nodes. The cropped graph is a combined graph of list of feature graphs.

featurebyte/query_graph/graph.py
```diff
@@ -281,6 +287,27 @@ def prune(self, target_node: Node, aggressive: bool) -> GraphNodeNameMap:
...
         return pruned_graph, node_name_map
 
+    def crop(self, target_node_names: List[str]) -> GraphCroppingOutput:
+        """
+        Crop the query graph and return the cropped graph & mapped node.
+
+        To crop the graph, this function first traverses from the target node to the input node.
+        The unused branches of the graph will get pruned in this step. After that, a new graph is
+        reconstructed by adding the required nodes back. Crop is less expensive than prune as it
+        does not extract the operation structure.
```
YS-L, line 297: From caller's perspective, is cropping a graph equivalent to non-aggressive pruning for a list of target node (aside from whether the operation structure is extracted internally)?

When doing non-aggressive pruning, would it make sense to use crop if it is less expensive?

gee-senbong: To make it clearer, i am thinking to change all the non-aggressive pruning from the client side to `crop` and `prune` should only have aggressive mode (aggressive mode parameter will be removed). Note that the amount of changes could be quite big if we make it in the same PR.

featurebyte/query_graph/model/graph.py
```diff
@@ -540,3 +539,4 @@ def add_operation(
 
 NodeNameMap = Dict[str, str]
 GraphNodeNameMap = Tuple[QueryGraphModel, NodeNameMap]
+GraphCroppingOutput = Tuple[QueryGraphModel, List[str]]
```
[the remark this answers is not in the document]
gee-senbong: 🆗 Will update this part to return the node mapping and handle the mapping outside the extractor.

featurebyte/query_graph/graph.py
YS-L: Got it, that makes sense to me.

Another minor comment is that I find the name `crop` unclear in conveying what the operation means and when to use it. Wondering if something like `quick_prune()` where input parameter is a list of nodes is clearer: it is non-aggressive pruning but faster and can work on a list of target nodes, and `prune()` is expected to be slower because it prunes more aggressively.

YS-L (approved):

tests/unit/query_graph/test_graph_quick_pruning.py
```diff
@@ -0,0 +1,77 @@
+"""
+Test for graph quick pruning related logics
+"""
+import pytest
+
+from featurebyte.query_graph.graph import QueryGraph
+
+
+def check_quick_pruned_graph(original_graph, quick_pruned_graph, target_nodes, mapped_node_names):
+    """Check the quick-pruned graph with non-aggressive pruned graph"""
```
YS-L, line 10: Is this test still valid now that pruning is always aggressive?

YS-L (approved):
Removing the aggressive parameter does look cleaner. LGTM

tests/unit/query_graph/test_graph_quick_pruning.py
gee-senbong: Yes, the test is still valid (but the doc string is confusing). It is comparing `original_graph.prune` output with `quick_pruned_graph.prune` output. Just to check the quick prune graph does not change anything unexpectedly.

PR #1426, opened 2023-06-14
[DEV-1788] api/target: add basic API object for Target

jevonyeoh:
## Description
This PR adds the basic API object for working with a `Target`. This allows users to create a `Target` via the constructor in the form of

```
fraud_flag_7d = Target.create(
  name='Fraudulent CC Transaction after 7 days',
  entities='CC Transaction',
  horizon="7d",
  blind_spot="1d",
)
```

This will also save the `Target`.

Will add more functionality (eg. `series` like functionality) in the future once we get the forward aggregates working.

## Related Issue
https://featurebyte.atlassian.net/browse/DEV-1788

## Type of Change

## Checklist

### DIFF
--- a/featurebyte/api/target.py
+++ b/featurebyte/api/target.py
@@ -1,0 +1,53 @@
+"""
+Target API object
+"""
+from typing import Any, List, Optional
+
+from pydantic import Field
+
+from featurebyte import Entity
+from featurebyte.api.api_object import ForeignKeyMapping
+from featurebyte.api.savable_api_object import SavableApiObject
+from featurebyte.models.base import PydanticObjectId
+from featurebyte.models.target import TargetModel
+from featurebyte.schema.target import TargetUpdate
+
+
+class Target(SavableApiObject):
+    """
+    Target class used to represent a Target in FeatureByte.
+    """
+
+    # Fields that a Target can be created with
+    internal_entity_ids: Optional[List[PydanticObjectId]] = Field(alias="entity_ids")
+    horizon: str
+    blind_spot: str
+
+    _route = "/target"
+    _update_schema_class = TargetUpdate
+
+    _list_schema = TargetModel
+    _get_schema = TargetModel
+    _list_fields = ["name", "entities"]
+    _list_foreign_keys = [
+        ForeignKeyMapping("entity_ids", Entity, "entities"),
+    ]
+
+    @property
+    def entities(self) -> List[Entity]:
+        """
+        Returns a list of entities associated with this target.
+
+        Returns
+        -------
+        List[Entity]
+        """
+        return [Entity.get_by_id(entity_id) for entity_id in self.internal_entity_ids]
+
+    def __init__(
+        self, name: str, entities: Optional[List[str]], horizon: str, blind_spot: str, **kwargs: Any
+    ):
+        entity_ids = [Entity.get(entity_name).id for entity_name in entities]
+        super().__init__(
+            name=name, entity_ids=entity_ids, horizon=horizon, blind_spot=blind_spot, **kwargs
+        )
--- a/featurebyte/api/target.py
+++ b/featurebyte/api/target.py
@@ -45,7 +45,7 @@ class Target(SavableApiObject):
         List[Entity]
         """
         try:
-            entity_ids = self.cached_model.entity_ids  # pylint: disable=attr-defined
+            entity_ids = self.cached_model.entity_ids  # type: ignore[attr-defined]
         except RecordRetrievalException:
             entity_ids = self.internal_entity_ids
         return [Entity.get_by_id(entity_id) for entity_id in entity_ids]
--- a/featurebyte/api/target.py
+++ b/featurebyte/api/target.py
@@ -42,12 +42,35 @@ class Target(SavableApiObject):
         -------
         List[Entity]
         """
+        if not self.internal_entity_ids:
+            return []
         return [Entity.get_by_id(entity_id) for entity_id in self.internal_entity_ids]
 
     def __init__(
         self, name: str, entities: Optional[List[str]], horizon: str, blind_spot: str, **kwargs: Any
     ):
-        entity_ids = [Entity.get(entity_name).id for entity_name in entities]
+        internal_kwargs = kwargs.copy()
+        entity_ids = None
+        if "entity_ids" in internal_kwargs:
+            entity_ids = internal_kwargs.pop("entity_ids")
+        if entities:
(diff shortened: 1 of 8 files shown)


featurebyte/api/target.py
```diff
@@ -0,0 +1,82 @@
...
+            entity_ids = internal_kwargs.pop("entity_ids")
+        if entities:
+            entity_ids = [Entity.get(entity_name).id for entity_name in entities]
+        super().__init__(
+            name=name,
+            entity_ids=entity_ids,
+            horizon=horizon,
+            blind_spot=blind_spot,
+            **internal_kwargs,
+        )
```
gee-senbong, line 70: Should we use `def create` method like what we used in other places like `Entity`, `Catalog`, `*Table`?

featurebyte/api/target.py
```diff
@@ -0,0 +1,82 @@
...
+        entity_names = None
+        if self.internal_entity_ids:
+            entity_names = [
+                Entity.get_by_id(entity_id).name for entity_id in self.internal_entity_ids
+            ]
+        return {"entities": entity_names}
+
+    @classmethod
+    def _get_init_params(cls) -> dict[str, Any]:
+        return {"entities": None}
```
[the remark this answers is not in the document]
jevonyeoh: there are some errors if they're not added - you can try checking out the branch, removing them and running the tests! wasn't too sure how best to resolve them heh

featurebyte/api/target.py
jevonyeoh: hm good point, i guess `FeatureGroup` and `FeatureList` don't use `create`, but the constructor, because they can be manipulated before being saved. i don't think that'll be the case here so might be worth doing a `create` function. let me check w/ xavier

featurebyte/api/target.py
```diff
@@ -0,0 +1,82 @@
...
+
+
+class Target(SavableApiObject):
+    """
+    Target class used to represent a Target in FeatureByte.
+    """
+
+    internal_entity_ids: Optional[List[PydanticObjectId]] = Field(alias="entity_ids")
+    horizon: Optional[str]
+    blind_spot: Optional[str]
```
gee-senbong, line 25: Maybe we should use `StrictStr` here. This prevent the case where user do something like `horizon=1` (if `str` is used, 1 is converted to '1'. If `StrictStr` is used, it will trigger error if non-string type is provided).

jevonyeoh: good point, will update.

featurebyte/api/target.py
```diff
@@ -0,0 +1,82 @@
...
+        """
+        Returns a list of entities associated with this target.
+
+        Returns
+        -------
+        List[Entity]
+        """
+        if not self.internal_entity_ids:
+            return []
+        return [Entity.get_by_id(entity_id) for entity_id in self.internal_entity_ids]
```
gee-senbong, line 48: We should use the pattern similar to https://github.com/featurebyte/featurebyte/blob/6e4db13c617a4b234f5efde2b6967137ee3335f1/featurebyte/api/event_table.py#L217-L229

The main idea is that we use `entity_ids` from the persistent object if it exists. Otherwise, use the one stored at the memory.

featurebyte/api/target.py
gee-senbong: 🆗 Got it.

featurebyte/api/target.py
```diff
@@ -0,0 +1,85 @@
+"""
+Target API object
+"""
+from __future__ import annotations
+
+from typing import Any, List, Optional
+
+from pydantic import Field, StrictStr
+
+from featurebyte import RecordRetrievalException
```
gee-senbong: Should use full import path here.

featurebyte/api/target.py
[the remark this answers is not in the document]
gee-senbong: It seems working to me. 🤔  Any particular tests?

featurebyte/api/target.py
gee-senbong: Currently we mostly used at API object layer & public api payload check.

featurebyte/api/target.py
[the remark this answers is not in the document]
jevonyeoh: o_O weird. i swear they weren't working yesterday. oh well! 

also xavier said `create` is good, so i'll update to that as well!

featurebyte/api/target.py
```diff
@@ -0,0 +1,90 @@
...
+        List[Entity]
+        """
+        try:
+            entity_ids = self.cached_model.entity_ids  # type: ignore[attr-defined]
+        except RecordRetrievalException:
+            entity_ids = self.internal_entity_ids
+        return [Entity.get_by_id(entity_id) for entity_id in entity_ids]
+
+    @classmethod
+    def create(
```
gee-senbong, line 54: Should we add `@typechecked` here?

featurebyte/models/target.py
```diff
@@ -33,17 +33,17 @@ class TargetModel(FeatureByteCatalogBaseDocumentModel):
     """
 
     # Recipe
-    graph: QueryGraph = Field(allow_mutation=False)
-    node_name: str
+    graph: Optional[QueryGraph] = Field(allow_mutation=False)
+    node_name: Optional[str] = Field(allow_mutation=False)
```
gee-senbong, line 37: Maybe we should move the `Field(allow_mutation=False)` to the api object. The reason that you may find it in our model file is because previously we inherit it from model. For our current api object, we try not to inherit document model directly.

featurebyte/api/target.py
```diff
@@ -0,0 +1,90 @@
...
+
+
+class Target(SavableApiObject):
+    """
+    Target class used to represent a Target in FeatureByte.
+    """
+
+    internal_entity_ids: Optional[List[PydanticObjectId]] = Field(alias="entity_ids")
+    horizon: Optional[StrictStr]
+    blind_spot: Optional[StrictStr]
```
gee-senbong, line 26: We should make `horizon` & `blind_spot` to following `internal_*` pattern too if we allow user to update it.

featurebyte/models/target.py
jevonyeoh: hm ok, so no `allow_mutation`s in the models?

gee-senbong (approved):
Left some non-critical comments. In general, LGTM.

featurebyte/models/target.py
gee-senbong: It is ok to keep it in the models. The main propose is to make it immutable for api object.

PR #1427, opened 2023-06-14
[chore][poc] api_object/list: refactor list to a list handler

jevonyeoh:
## Description
This PR is a proof-of-concept for a refactor to extract list logic out of the `ApiObject` class, and into a separate handler.

Specifically, we also
- move `@staticmethods` out of the class into `api_object_util`: benefit of this is that it's clearer that there are no dependencies on the class, and that these functions do not depend on any state. This makes it clearer when we want to reuse them.
- encapsulate all the `list` function dependencies (eg. route, schema, fields, and foreign_key mappings) into the new handler object: benefit here is that it's clearer to a new reader what class variables are actually used to configure the `list` operation
- encapsulate list functionality into the `list_handler` object: benefit is that it's easier to unit test this functionality without having to try to create an `ApiObject` itself. 

## Type of Change

## Checklist

### DIFF
--- a/featurebyte/api/api_handler/list.py
+++ b/featurebyte/api/api_handler/list.py
@@ -9,6 +9,7 @@ from itertools import groupby
 from pandas import DataFrame
 
 from featurebyte.api.api_object_util import (
+    PAGINATED_CALL_PAGE_SIZE,
     ForeignKeyMapping,
     iterate_api_object_using_paginated_routes,
     map_dict_list_to_name,
@@ -16,11 +17,13 @@ from featurebyte.api.api_object_util import (
 )
 from featurebyte.models.base import FeatureByteBaseDocumentModel
 
-# TODO: consolidate this with the one in api_object.py
-PAGINATED_CALL_PAGE_SIZE = 100
-
 
 class ListHandler:
+    """
+    ListHandler is a class that handles listing of objects. ApiObject classes can use this class to configure how
+    to list objects.
+    """
+
     def __init__(
         self,
         route: str,
@@ -28,6 +31,18 @@ class ListHandler:
         list_fields: List[str] = ("name", "created_at"),
         list_foreign_keys: List[ForeignKeyMapping] = (),
     ):
+        """
+        Parameters
+        ----------
+        route: str
+            Route to list objects
+        list_schema: Type[FeatureByteBaseDocumentModel]
+            Schema to use to parse list response
+        list_fields: List[str]
+            Fields to include in list
+        list_foreign_keys: List[ForeignKeyMapping]
+            List of foreign keys that can be mapped to an object name
+        """
         self.route = route
         self.list_schema = list_schema
         self.list_fields = list_fields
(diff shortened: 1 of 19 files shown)


featurebyte/api/api_handler/catalog_list_handler.py
```diff
@@ -0,0 +1,18 @@
...
+from featurebyte.api.api_handler.list import ListHandler
+from featurebyte.models.base import get_active_catalog_id
+
+
+class CatalogListHandler(ListHandler):
+    """
+    Additional handling for catalog.
+    """
+
+    def additional_post_processing(self, item_list: pd.DataFrame) -> pd.DataFrame:
```
jevonyeoh, line 15: an example if an api object needs to perform additional post processing

featurebyte/api/api_object.py
```diff
@@ -441,6 +367,35 @@ def list(cls, include_id: Optional[bool] = True) -> DataFrame:
...
+        """
+        return ListHandler(
+            route=cls._route,
+            list_schema=cls._list_schema,
+            list_fields=cls._list_fields,
+            list_foreign_keys=cls._list_foreign_keys,
+        )
+
+    @classmethod
+    def use_new_list_handler(cls) -> bool:
```
jevonyeoh, line 388: temporary class method for migrating the code paths

featurebyte/api/api_object_util.py
```diff
@@ -128,3 +159,113 @@ def is_server_mode() -> bool:
...
+    if isinstance(object_dict, list):
+        return [
+            object_map.get(_obj_dict.get(object_id_field))
+            for _obj_dict in object_dict
+            if _obj_dict.get(object_id_field)
+        ]
+    return object_map.get(object_dict.get(object_id_field))
+
+
+def to_request_func(response_dict: dict[str, Any], page: int) -> bool:
```
gee-senbong, line 216: nit: We could make this function private as it is not used outside this module.

featurebyte/api/api_handler/list.py
```diff
@@ -0,0 +1,162 @@
...
+                    foreign_key_field = foreign_key_mapping.foreign_key_field
+                    if "." in foreign_key_field:
+                        # foreign_key is a dict
+                        foreign_key_field, object_id_field = foreign_key_field.split(".")
+                        mapping_function = partial(
+                            map_dict_list_to_name, object_map, object_id_field
+                        )
+                    else:
+                        # foreign_key is an objectid
+                        mapping_function = partial(map_object_id_to_name, object_map)
```
gee-senbong, line 141: Would it be cleaner to move the both `map_dict_list_to_name` & `map_object_id_to_name` to this class as they are not used elsewhere?

gee-senbong (approved):
LGTM. As our current API object listing logics become more & more complicated, it is good to have a separate class to encapsulate the logic.

featurebyte/api/api_object_util.py
jevonyeoh: will update in follow up :)

PR #1434, opened 2023-06-16
[DEV-1789] forward_aggregator: add ForwardAggregator class

jevonyeoh:
## Description
Implement the `ForwardAggregator` class, which will add a `FORWARD_AGGREGATE` node into the graph when it's called. 

This currently isn't called anywhere, but will be added into the `GroupBy` class shortly. 

Also makes a minor refactor to move the `BaseAggregator` into a separate `api/aggregator` folder.

## Type of Change

## Checklist


featurebyte/query_graph/node/generic.py
```diff
@@ -468,6 +469,56 @@ def _derive_sdk_code(
...
+        horizon: Optional[str]
+        blind_spot: Optional[str]
+        keys: List[InColumnStr]  # group by keys
+        parent: Optional[InColumnStr]
+        agg_func: AggFunc
+        value_by: InColumnStr  # aggregate column
+        serving_name: str
+        entity_ids: Optional[List[PydanticObjectId]]
+        # TableDetails are needed because we will be performing the aggregation query directly on the calling table.
+        table_details: TableDetails
```
gee-senbong, line 491: Should we inherit `featurebyte.query_graph.node.mixin.BaseGroupbyParameters`? If not, we need to update the logic in entity extractor to consider target  (https://github.com/featurebyte/featurebyte/blob/main/featurebyte/query_graph/transform/entity_extractor.py#L34-L49). Entity extractor is used to extract entity(s) of the feature (i suspect we likely need to reuse the same logic for target).

tests/unit/api/aggregator/test_forward_aggregator.py
```diff
@@ -0,0 +1,126 @@
...
+                target_name=target_name,
+            )
+    else:
+        forward_aggregator._validate_parameters(
+            value_column=value_column,
+            method=method,
+            horizon=horizon,
+            blind_spot=blind_spot,
+            target_name=target_name,
+        )
```
gee-senbong, line 126: Do we have a test to check whether the node output category is derived properly? This is to cover the changes in `NodeOutputCategory`.

featurebyte/api/aggregator/forward_aggregator.py
```diff
@@ -0,0 +1,176 @@
...
+    def forward_aggregate(
+        self,
+        value_column: str,
+        method: Optional[str] = None,
+        horizon: Optional[str] = None,
+        blind_spot: Optional[str] = None,
+        target_name: Optional[str] = None,
+    ) -> Target:
+        """
+        Aggregate given value_column for each group specified in keys over a list of time windows.
```
YS-L, line 38: This should be a single window. Would it be clearer to mention horizon instead of window in the docstring?

featurebyte/api/aggregator/forward_aggregator.py
```diff
@@ -0,0 +1,176 @@
...
+
+        Parameters
+        ----------
+        value_column: str
+            Column to be aggregated
```
YS-L, line 43: Should the type of this be None? `_validate_method_and_value_column` enforces that value column is None for count method.

featurebyte/api/aggregator/forward_aggregator.py
```diff
@@ -0,0 +1,176 @@
...
+            method=method,
+            horizon=horizon,
+            blind_spot=blind_spot,
+            target_name=target_name,
+        )
+        # Add forward aggregate node to graph.
+        forward_aggregate_node = self.view.graph.add_operation(
+            node_type=NodeType.FORWARD_AGGREGATE,
+            node_params=node_params,
+            node_output_type=NodeOutputType.SERIES,
```
YS-L, line 77: I think for consistency the output type should be frame.

tests/unit/api/aggregator/test_forward_aggregator.py
jevonyeoh: good point - added a simple test to extract operation structure, and verified the `output_category` is `target`!

featurebyte/query_graph/node/generic.py
jevonyeoh: also inherited the `AggregationOpStructMixin` to avoid having to duplicate the derive operation structure info if that seems reasonable!

YS-L (approved):

PR #1439, opened 2023-06-17
enhancement: add feature list with batch feature creation async route

gee-senbong:
## Description

This PR introduces a new feature list creation (with batch features creation) route to reduce number of API calls required for existing feature list creation route.

Existing SDK feature list creation will do the following:
* check whether feature in the feature list has been saved or not (take n number of api calls, n is the number of features in the feature list).
* post batch feature creation async task
* post feature list creation route

For the new flow, following will happen:
* post feature list with batch feature creation async task (inside the task, it will check whether the feature has been saved, perform batch feature creation, and then create the new feature list).

## Type of Change

## Checklist


featurebyte/api/feature_group.py
```diff
@@ -495,9 +483,31 @@ def save(self, conflict_resolution: ConflictResolution = "raise") -> None:
...
+        )
+
+        # save temporary feature list with list of features
+        self.post_async_task(
+            route="/feature_list/batch",
+            payload=feature_list_batch_feature_create.json_dict(),
+            retrieve_result=False,
+            has_output_url=False,
+        )
+        feature_list_dict = get_api_object_by_id(
```
YS-L, line 500: Would it make sense to wrap the lines below in a try block and delete the temporary feature list in the finally section to ensure clean up on unexpected errors?

gee-senbong: The temporary feature list is only created if all of the features have been created successfully. It is the last step of the async job. If we put it inside the try-finally block, it will becomes something like:
```
try:
    ....
finally:
    try:
         delete_api_object_by_id(route="/feature_list", id_value=feature_list_batch_feature_create.id)
    except:
         pass
```

gee-senbong: It may still have some value if there are some unexpected error happens in between finished async job & deletion happens, though i think the chance is small.

featurebyte/routes/feature/controller.py
```diff
@@ -96,9 +96,13 @@ async def submit_batch_feature_create_task(self, data: BatchFeatureCreate) -> Op
         Optional[Task]
             Task object
         """
+        # as there is no direct way to get the conflict resolved feature id for batch feature creation task,
+        # the conflict resolution should only support "raise" for public API. Therefore, we should not include
+        # the conflict resolution in the API payload schema (BatchFeatureCreate).
```
YS-L, line 101: Does this affect whether we can support "retrieve" resolution in any of the SDK feature / feature list creation methods?

gee-senbong: Technically, both feature & feature list async task can support `retrieve` resolution. The problem is that there is no direct way for feature save to retrieve resolved id. That's why `conflict_resolution` parameter is excluded from batch feature creation POST route. For feature list with batch feature creation task, we can use temporary feature list to obtain resolved feature id. That's why `conflict_resolution` parameter is kept for feature list with batch feature creation POST route.

PR #1443, opened 2023-06-19
[DEV-1824] forward_aggregator: generate SQL code for forward aggregator node

jevonyeoh:
## Description
Add logic to generate SQL code for the forward aggregator node.

## Related Issue
https://featurebyte.atlassian.net/browse/DEV-1824

## Type of Change

## Checklist


featurebyte/query_graph/sql/aggregator/forward.py
```diff
@@ -0,0 +1,155 @@
...
+                name=spec.parameters.value_by,
+            )
+            if spec.parameters.value_by
+            else None
+        )
+        groupby_columns = [
+            GroupbyColumn(
+                agg_func=s.parameters.agg_func,
+                parent_expr=(
+                    get_qualified_column_identifier(s.parameters.parent, "TABLE")
```
jevonyeoh, line 109: is the `TABLE` string arbitrary? what should it match?

featurebyte/query_graph/sql/specs.py
```diff
@@ -698,6 +701,47 @@ def construct_specs(
...
+
+    def agg_result_name(self) -> str:
+        return self.get_agg_result_name_from_groupby_parameters(self.parameters)
+
+    @property
+    def aggregation_type(self) -> AggregationType:
+        return AggregationType.FORWARD
+
+    def get_source_hash_parameters(self) -> dict[str, Any]:
+        # TODO:
```
jevonyeoh, line 720: wasn't sure what else needed to be added here

tests/unit/query_graph/sql/aggregator/test_forward.py
```diff
@@ -0,0 +1,103 @@
...
+
+
+@pytest.fixture(name="forward_node_parameters")
+def forward_node_parameters_fixture(entity_id):
+    return ForwardAggregateParameters(
+        name="target",
+        timestamp_col="timestamp_col",
+        horizon="7d",
+        blind_spot="1d",
+        table_details=TableDetails(table_name="table"),
```
jevonyeoh, line 23: how do we utilize the table details in the sql generation?

featurebyte/query_graph/sql/aggregator/forward.py
```diff
@@ -0,0 +1,155 @@
...
+        ]
+        groupby_input_expr = (
+            select()
+            .from_(
+                expressions.Table(
+                    this=quoted_identifier(self.request_table_plan.get_request_table_name(spec)),
+                    alias="REQ",
+                )
+            )
+            .join(point_in_time_expr, join_type="inner", on=record_validity_condition)
```
YS-L, line 125: I think this should be joining with `spec.source_expr.subquery(alias="TABLE")` instead of `point_in_time_expr`. Using the `TABLE` alias here to match with the expressions defined above.

featurebyte/query_graph/sql/specs.py
YS-L: For now this can follow what was done for item aggregate: https://github.com/featurebyte/featurebyte/blob/5fab5ef8cb45b1bd94686ecf046fcc5d551f5e5b/featurebyte/query_graph/sql/specs.py#L527-L535

The parameters here are used to determine the specs that can be processed in the same subquery. Should be possible to refactor and avoid duplicated logic later.

YS-L (commented):
Is it possible to add a test case that generates the sql through a `FeatureExecutionPlanner`? Similar tests can be found in `featurebyte/query_graph/sql/feature_preview.py`.

I think this might be missing adding a corresponding class in `featurebyte/query_graph/sql/ast/aggregate.py`. The test above should help catch what is missing.

jevonyeoh:
hm @YS-L - i added a test in `query_graph/sql/test_forward_aggregate` that calls into a method that utilizes a `FeatureExecutionPlanner`. i'm not sure why the generated SQL still has tile-related queries though 🤔

featurebyte/query_graph/sql/feature_compute.py
```diff
@@ -68,6 +74,7 @@ def __init__(
             AggregationType.WINDOW: WindowAggregator(**aggregator_kwargs),
             AggregationType.ITEM: ItemAggregator(**aggregator_kwargs),
             AggregationType.AS_AT: AsAtAggregator(**aggregator_kwargs),
+            AggregationType.FORWARD: ForwardAggregator(**aggregator_kwargs),
```
YS-L, line 77: It looks like the forward aggregate specs are not being picked up. I think we need to add a condition here in `FeatureExecutionPlanner` to handle the new aggregation type:

https://github.com/featurebyte/featurebyte/blob/3f8ed5967b3362a0bd4771b71689cee8695bf414/featurebyte/query_graph/sql/feature_compute.py#L470-L488

tests/unit/query_graph/sql/test_forward_aggregate.py
```diff
@@ -0,0 +1,28 @@
+"""
+Test target forward aggregate module
+"""
+from featurebyte import SourceType
+from featurebyte.query_graph.sql.common import REQUEST_TABLE_NAME
+from featurebyte.query_graph.sql.forward_aggregate import get_forward_aggregate_sql
+from tests.util.helper import assert_equal_with_expected_fixture
+
+
+def test_get_forward_aggregate_sql(query_graph_with_groupby, update_fixtures):
```
YS-L, line 10: `query_graph_with_groupby` contains normal window aggregates and that is probably why the generated sql is still showing tile related queries. I think we should add a new fixture in `tests/unit/query_graph/conftest.py` with a forward aggregate node.

tests/unit/query_graph/sql/aggregator/test_forward.py
```diff
@@ -0,0 +1,109 @@
...
+              TABLE."col_float" AS "col_float",
+              SUM(TABLE."value") AS "_fb_internal_forward_sum_value_cust_id_col_float_input_1_inner"
+            FROM "REQUEST_TABLE_POINT_IN_TIME_serving_cust_id" AS REQ
+            INNER JOIN (
+              SELECT
+                *
+              FROM tab
+            ) AS TABLE
+              ON TABLE."timestamp_col" > REQ."POINT_IN_TIME"
+              AND TABLE."timestamp_col" <= FLOOR(DATE_PART(EPOCH_SECOND, REQ."POINT_IN_TIME") + 604800.0)
```
YS-L, line 95: Should we convert the left hand side `TABLE."timestamp_col"` to epoch seconds too before comparing?

featurebyte/query_graph/sql/aggregator/forward.py
```diff
@@ -0,0 +1,164 @@
...
+            .from_(
+                expressions.Table(
+                    this=quoted_identifier(self.request_table_plan.get_request_table_name(spec)),
+                    alias="REQ",
+                )
+            )
+            .join(
+                spec.source_expr.subquery(alias="TABLE"),
+                join_type="inner",
+                on=record_validity_condition,
```
YS-L, line 133: The join condition should also include equality on join keys: `REQ.{serving_name} = TABLE.{key}` (could be a list).

tests/unit/query_graph/conftest.py
```diff
@@ -357,6 +357,29 @@ def query_graph_with_groupby_fixture(query_graph_and_assign_node, groupby_node_p
...
 
 
+@pytest.fixture(name="query_graph_with_forward_aggregate")
+def query_graph_with_forward_aggregate_fixture(query_graph_and_assign_node, event_table_details):
+    """Fixture of a query graph with a forward aggregate operation"""
+    graph, assign_node = query_graph_and_assign_node
+    graph.add_operation(
+        node_type=NodeType.FORWARD_AGGREGATE,
+        node_params={
+            "name": "target_node",
```
YS-L, line 367: nit: Might be a bit less confusing to use an actual target name here because it is unrelated to graph or nodes.

featurebyte/query_graph/sql/aggregator/forward.py
```diff
@@ -0,0 +1,170 @@
...
+                    get_qualified_column_identifier(s.parameters.parent, "TABLE")
+                    if s.parameters.parent
+                    else None
+                ),
+                result_name=s.agg_result_name,
+            )
+            for s in specs
+        ]
+        serving_name = quoted_identifier(spec.serving_names[0]).sql()
+        join_table_key = quoted_identifier(spec.parameters.keys[0]).sql()
```
YS-L, line 123: There could be multiple join keys and in that case they should be `and_`'ed.

YS-L (approved):
This looks good overall. Would be good to add an integration test similar to `tests/integration/api/test_feature_correctness.py` to ensure that the generated sql is correct in this or a next PR.

featurebyte/query_graph/sql/aggregator/forward.py
jevonyeoh: hm i assumed that targets will only have one serving name / join key. is that different?

YS-L: In the groupby method the join key can be a list. The forward aggregate is called through the same groupby method right?

jevonyeoh: oh yes, sorry i got confused. i'll fix this in a follow-up since the tests here are almost done 😁

PR #1462, opened 2023-06-28
[chore] data_migrations: update to use injection via config

jevonyeoh:
## Description
This should make it easier to reuse services across the migration framework too.

Will also make it less annoying to add dependencies to existing services that might be used in these migration services.

Changed to traditional constructors so that it's easier to see what the dependencies are in each of the classes, instead of having them declared all over the place.

## Type of Change

## Checklist

### DIFF
--- a/featurebyte/migration/service/mixin.py
+++ b/featurebyte/migration/service/mixin.py
@@ -173,7 +173,6 @@ class DataWarehouseMigrationMixin(BaseMigrationServiceMixin, ABC):
         super().__init__(persistent=persistent)
         self.session_manager_service = session_manager_service
         self.feature_store_service = feature_store_service
-        self._allow_to_use_raw_query_filter = False
 
     @property
     def collection_name(self) -> str:
@@ -193,7 +192,7 @@ class DataWarehouseMigrationMixin(BaseMigrationServiceMixin, ABC):
 
     @contextmanager
     def allow_use_raw_query_filter(self) -> Iterator[None]:
-        return self.feature_store_service.allow_use_raw_query_filter()
+        yield self.feature_store_service.allow_use_raw_query_filter()
 
     async def get_session(self, feature_store: FeatureStoreModel) -> BaseSession:
         """
--- a/featurebyte/migration/service/mixin.py
+++ b/featurebyte/migration/service/mixin.py
@@ -148,10 +148,8 @@ class DataWarehouseMigrationMixin(BaseMigrationServiceMixin, ABC):
         # the feature store user ID instead of the arbitrary current user ID. This is required for these
         # services as we need to be able to retrieve the correct credentials from the persistent layer.
         # We can use an arbitrary user ID for other services as there's no dependency on user ID.
-        self.session_manager_service.user = feature_store_user
-        self.session_manager_service.session_validator_service.user = feature_store_user
         session = await self.session_manager_service.get_feature_store_session(
-            feature_store, get_credential=self.get_credential
+            feature_store, get_credential=self.get_credential, user_override=feature_store_user
         )
         return session
 
--- a/featurebyte/migration/service/mixin.py
+++ b/featurebyte/migration/service/mixin.py
@@ -14,7 +14,8 @@ from featurebyte.exception import CredentialsError
 from featurebyte.logging import get_logger
 from featurebyte.models.base import FeatureByteBaseDocumentModel, FeatureByteBaseModel
 from featurebyte.models.feature_store import FeatureStoreModel
-from featurebyte.persistent.base import Document, Persistent
+from featurebyte.models.persistent import Document
+from featurebyte.persistent.base import Persistent
 from featurebyte.schema.common.base import BaseDocumentServiceUpdateSchema
 from featurebyte.service.base_document import BaseDocumentService
 from featurebyte.service.feature_store import FeatureStoreService
--- a/featurebyte/migration/service/mixin.py
+++ b/featurebyte/migration/service/mixin.py
@@ -44,7 +44,7 @@ class BaseMigrationServiceMixin:
         """
 
     @abstractmethod
-    def _construct_list_query_filter(
+    def construct_list_query_filter(
         self,
         query_filter: Optional[dict[str, Any]] = None,
         use_raw_query_filter: bool = False,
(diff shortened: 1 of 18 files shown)


featurebyte/migration/service/mixin.py
```diff
@@ -183,19 +143,7 @@ async def get_session(self, feature_store: FeatureStoreModel) -> BaseSession:
         -------
         BaseSession
         """
-        credential_provider = MongoBackedCredentialProvider(persistent=self.persistent)
-        user = User(id=feature_store.user_id)
```
YS-L, line 187: I think it is important that the `user` here is the feature store's user since the session services below need to be specific to that user. See this PR: https://github.com/featurebyte/featurebyte/pull/681

If I recall correctly, the user id that gets passed when calling `run_migration()` is a dummy value.

@gee-senbong could you confirm?

tests/unit/migration/test_run.py
```diff
@@ -223,21 +233,22 @@ async def test_run_migration(migration_check_persistent, user, get_credential):
...
     mock_session_manager,
     migration_check_persistent,
-    user,
     test_dir,
+    user,
+    app_container,
 ):
     """Test data warehouse migration get_session method"""
     fixture_path = os.path.join(test_dir, "fixtures/request_payloads/feature_store.json")
-    feature_store_user_id = ObjectId()
```
YS-L, line 231: I think setting the feature store user id to a different value than `user.id` was the intention of this test.

featurebyte/migration/service/mixin.py
gee-senbong: YS is right. We need to use the feature store's `user_id` to construct the session. Otherwise, it will cause issue when connecting to the warehouse during migration.

tests/unit/migration/test_run.py
gee-senbong: Yes. That is the purpose of this test. I should add more comment on this test. 😅

featurebyte/migration/service/mixin.py
jevonyeoh: ah ok! let me fix this

jevonyeoh: just to double check - this is the entry point where the user id being passed in is the dummy one yea @gee-senbong / @YS-L 

https://github.com/featurebyte/featurebyte/blob/27032840d56d21add4e597b3b1569588316d46f5/docker/migration.py#L19-L21

YS-L: Yes, that's right. There is a similar entry point in the featurebyte-app as well.

jevonyeoh: hmm.. ok, i'm going to have to sit on this a bit more.

jevonyeoh: curious - why do we even pass in a user here into `run_migration` if it's going to be updated at at later point?

i see that we initialize a `SchemaMetadataService` with the arbitrary user id - i guess that's ok?
https://github.com/featurebyte/featurebyte/blob/27032840d56d21add4e597b3b1569588316d46f5/featurebyte/migration/run.py#L210-L215

jevonyeoh: i added some special handling for these services here! @YS-L / @gee-senbong

featurebyte/migration/service/mixin.py
```diff
@@ -183,19 +143,14 @@ async def get_session(self, feature_store: FeatureStoreModel) -> BaseSession:
...
-        )
-        session = await session_manager_service.get_feature_store_session(
+        feature_store_user = User(id=feature_store.user_id)
+        # A hack to update the user ID in the session manager service and session validator service to use
+        # the feature store user ID instead of the arbitrary current user ID. This is required for these
+        # services as we need to be able to retrieve the correct credentials from the persistent layer.
+        # We can use an arbitrary user ID for other services as there's no dependency on user ID.
+        self.session_manager_service.user = feature_store_user
+        self.session_manager_service.session_validator_service.user = feature_store_user
+        session = await self.session_manager_service.get_feature_store_session(
```
gee-senbong, line 153: @jevonyeoh Could we add a parameter in `self.session_manager_service.get_feature_store_session` to retrieve the credential based on provided `user_id`? If we override hte `user` here, not sure whether it will cause other issue if we run this on multiple threads.

@YS-L Once your PR (https://github.com/featurebyte/featurebyte/pull/1476) get merged, my understanding is that we don't need to migrate feature store metadata, right?

jevonyeoh: good idea - updated!

gee-senbong: Maybe we could remove the comment above.

tests/unit/migration/test_run.py
gee-senbong: 👍 For adding comment

tests/unit/routes/test_lazy_app_container.py
```diff
@@ -121,7 +121,7 @@ def test_construction__empty_app_config_has_two_instances(app_container_construc
         **app_container_constructor_params,
         app_container_config=app_container_config,
     )
-    assert len(app_container.instance_map) == 4
+    assert len(app_container.instance_map) == 5
```
gee-senbong, line 125: If i understood this correctly, this test `test_construction__empty_app_config_has_two_instances` is to check how many instances within the empty app, right?

gee-senbong (approved):
LGTM. Thanks for the update 👍

tests/unit/routes/test_lazy_app_container.py
jevonyeoh: yeap! this is updated because i added the `persistent` object as one of the default ones in the empty app!

gee-senbong: Ok. The `two_instances` in the test name seems a bit confusing to me.

jevonyeoh: ah i probably never updated the name. i'll be making another slight tweak to this shortly, so i'll update it in a follow-up.

featurebyte/migration/service/mixin.py
YS-L: While this solution works for now, I'm wondering if later similar issues will arise for other services as well. Adding user override all over the places seems not ideal. When that need arises, I think we have to revisit the way the services are instantiated within the migration services.

featurebyte/service/working_schema.py
```diff
@@ -54,54 +46,21 @@ class WorkingSchemaService(BaseService):
...
-        self.feature_manager_service = FeatureManagerService(
-            user=user,
-            persistent=persistent,
-            catalog_id=catalog_id,
-            tile_manager_service=self.tile_manager_service,
-        )
+        self.tile_scheduler_service = tile_scheduler_service
+        self.tile_manager_service = tile_manager_service
+        self.feature_manager_service = feature_manager_service
+        self.feature_service = feature_service
```
YS-L, line 63: All these services are now targeting a dummy user id. Just making a note here, will need to double check whether that will cause any issues when the services are used.

YS-L (approved):

PR #1463, opened 2023-06-30
feature: add generic function node

gee-senbong:
## Description

This PR implements generic function node (used by query graph) to support user-defined-function registered at the data warehouse.

## Type of Change

## Checklist


featurebyte/query_graph/node/function.py
```diff
@@ -0,0 +1,151 @@
...
+    CodeGenerationContext,
+    ExpressionStr,
+    StatementT,
+    ValueStr,
+    VariableNameGenerator,
+    VarNameExpressionInfo,
+)
+
+
+class FunctionParameterInput(BaseModel):
```
jevonyeoh, line 32: does it make sense to have something like 

```
class BaseFunctionParameter:
  dtype: DBVarType
  def get_column_args -> Optional[str]:
    return None
  def get_literal_value -> expressions.Expression:
    # get each implementation to override
    pass

class ValueFunctionParameter(BaseFunctionParameter)
  value: Any

class ColumnFunctionParameter(BaseFunctionParameter)
  column: str
  def get_column_args -> Optional[str]:
    return self.column
```

this way, we should be able to skip all the if-else checks on the `input_form` type

gee-senbong: Good point. If i understand this correctly, basically we will have different function parameter: column type & value type. Will update this accordingly.

featurebyte/query_graph/node/function.py
```diff
@@ -0,0 +1,227 @@
...
+        operation_structure: OperationStructure,
+        config: CodeGenerationConfig,
+        context: CodeGenerationContext,
+    ) -> Tuple[List[StatementT], VarNameExpressionInfo]:
+        function_parameters: List[Any] = []
+        node_input_index = 0
+        for func_param in self.parameters.function_parameters:
+            func_param_val, node_input_index = func_param.get_sdk_function_argument(
+                node_inputs, node_input_index
+            )
```
YS-L, line 222: I'm not sure if the `get_sdk_function_argument()` abstraction is necessary since `ValueFunctionParameterInput` basically ignores `node_inputs` and `node_input_index`.

Wondering if something simpler like this is easier to understand:

```python
if func_param.input_form == "value":
    func_param_val = ValueStr.create(func_param.value)
else:
    func_param_val = node_inputs[node_input_index]
    node_input_index += 1
```

gee-senbong: This is what i did initially and it is suggested by @jevonyeoh to abstract the logic here. @jevonyeoh

featurebyte/query_graph/node/function.py
```diff
@@ -0,0 +1,227 @@
...
+        -------
+        Tuple[RightHandSide, int]
+            Tuple of SDK function argument and the updated node input index.
+        """
+
+
+class ValueFunctionParameterInput(BaseFunctionParameterInput):
+    """ValueFunctionParameterInput class"""
+
+    value: Optional[Any]
```
YS-L, line 77: Would the type that we use for scalar parameters be appropriate here? https://github.com/featurebyte/featurebyte/blob/8377eb48f17d77f1f6cd5637b18074ec08a915bb/featurebyte/query_graph/node/scalar.py#L59

Seems better if we can avoid Any if not necessary.

gee-senbong: Will update this to use `ValueParameterType`

featurebyte/query_graph/node/function.py
```diff
@@ -0,0 +1,227 @@
...
+        self, node_inputs: List[VarNameExpressionInfo], node_input_index: int
+    ) -> Tuple[RightHandSide, int]:
+        # do not increment node_input_index as this is a value input, it should not increment the node_input_index
+        return ValueStr.create(self.value), node_input_index
+
+
+class ColumnFunctionParameterInput(BaseFunctionParameterInput):
+    """ColumnFunctionParameterInput class"""
+
+    column_name: Optional[str]
```
YS-L, line 95: Is the `column_name` optional because the input can be a temporary column that is not assigned into a view? If the input is a feature, will this refer to the feature name?

featurebyte/query_graph/node/function.py
```diff
@@ -0,0 +1,228 @@
...
+        self, node_inputs: List[VarNameExpressionInfo], node_input_index: int
+    ) -> Tuple[RightHandSide, int]:
+        # InfoDict is not expected here as it should be used only in (ConditionNode - AssignNode) structure
+        value = node_inputs[node_input_index]
+        assert not isinstance(value, InfoDict), "Unexpected InfoDict type"
+        # increment node_input_index as this is a column input, it should increment the node_input_index
+        return value, node_input_index + 1
+
+
+FunctionParameterInput = Annotated[
```
jevonyeoh, line 114: curious - is this type needed? could `function_parameters` be of the type `List[BaseFunctionParameterInput]` instead?

featurebyte/query_graph/node/function.py
jevonyeoh: i think it's easier to extend in the current form if we ever need to add more types of inputs since we can just add another implementation, and not have to worry about updating all the if-statements.

featurebyte/query_graph/node/function.py
gee-senbong: Yes, `column_name` because it can be a derived column if it is not assigned to a view. If the input is a feature, this will refer to the feature name.

featurebyte/query_graph/node/function.py
```diff
@@ -0,0 +1,228 @@
...
+from featurebyte.query_graph.node.scalar import ValueParameterType
+
+
+class BaseFunctionParameterInput(BaseModel):
+    """BaseFunctionParameterInput class"""
+
+    dtype: DBVarType
+
+    @abstractmethod
+    def get_column_args(self) -> List[Optional[str]]:
```
jevonyeoh, line 44: curious - when we would pass in `None` as a column arg?

featurebyte/query_graph/node/function.py
gee-senbong: The value of `column_name` is not used in SDK feature derivation. The main reason to keep this is to support `get_required_input_columns` method (https://github.com/featurebyte/featurebyte/blob/cc24823e663e72e1a9369785d33950c637ce57cd/featurebyte/query_graph/transform/reconstruction.py#L167-L172).

featurebyte/query_graph/node/function.py
gee-senbong: This is needed. Without this, the input parameter will be dropped after deserialization.

featurebyte/query_graph/node/function.py
gee-senbong: https://github.com/featurebyte/featurebyte/pull/1463#discussion_r1247488091

When we derivate a new column from existing column, the new column's name is `None`.

featurebyte/query_graph/node/function.py
jevonyeoh: got it, thanks!

featurebyte/query_graph/node/function.py
jevonyeoh: oh ok, just saw that!

featurebyte/query_graph/node/function.py
gee-senbong: For example, `BaseFunctionParameterInput` only contains `dtype`. Those additional attributes that are specific to value or column will be dropped after deserialization (since the expected type is `BaseFunctionParameterInput` and it does not contain these parameters).

featurebyte/query_graph/node/function.py
```diff
@@ -0,0 +1,228 @@
...
+            ]
+
+        return OperationStructure(
+            **node_kwargs,
+            output_type=NodeOutputType.SERIES,
+            output_category=input_category,
+            row_index_lineage=row_index_lineage,
+        )
+
+    def _derive_sdk_code(
```
YS-L, line 210: Will test be added to cover feature definition generation?

gee-senbong: Tests are added in `tests/unit/query_graph/test_function.py`.
```
    # check the code generated by the generic function node
    state = SDKCodeExtractor(graph=global_graph).extract(node=gfunc)
    assert state.code_generator.statements[-1] == ("output", 'my_func(col, 1, 2.0, "hello", True)')
```

YS-L (approved):

PR #1465, opened 2023-06-30
feature: add user defined function endpoints & tests

gee-senbong:
## Description

This PR introduces following user-defined-function endpoints:
* `POST /user_defined_function`
* `GET /user_defined_function`
* `GET /user_defined_function/<user_defined_function_id>`
* `PATCH /user_defined_function/<user_defined_function_id>`
* `DELETE /user_defined_function/<user_defined_function_id>`

## Type of Change

## Checklist


featurebyte/routes/user_defined_function/controller.py
```diff
@@ -0,0 +1,226 @@
...
+            # check if user defined function exists in warehouse
+            await self.feature_store_warehouse_service.check_user_defined_function_exists(
+                user_defined_function=user_defined_function,
+                feature_store=feature_store,
+                get_credential=get_credential,
+            )
+        except Exception as exc:
+            # if not exists, delete the user defined function
+            await self.service.delete_document(document_id=user_defined_function.id)
+            raise DocumentCreationError(f"User defined function not exists: {exc}") from exc
```
YS-L, line 91: If the function exists but `get_default_test_value()` raises ValueError because the parameter type is not supported, will it be raised through here? Would be a bit nicer if the error message can reflect that instead of always mentioning the function does not exist.

featurebyte/routes/user_defined_function/controller.py
```diff
@@ -0,0 +1,226 @@
...
+
+        func_param_mapping = OrderedDict()
+        for func_param in document.function_parameters:
+            func_param_mapping[func_param.name] = func_param
+
+        for func_param in data.function_parameters:
+            if func_param.name in func_param_mapping:
+                func_param_mapping[func_param.name] = func_param
+            else:
+                raise DocumentUpdateError(f"Function parameter not exists: {func_param.name}")
```
YS-L, line 141: Does this mean only the parameter types can be updated? Curious if there is a reason to be this strict because I imagine it would be useful if the function signature can be updated in case the function was registered incorrectly (and when it's not used by any features). Either that or don't support updating function signature at all makes more sense to me.

featurebyte/routes/user_defined_function/controller.py
```diff
@@ -0,0 +1,226 @@
...
+
+        function_parameters = list(func_param_mapping.values())
+        if function_parameters == document.function_parameters:
+            raise DocumentUpdateError("No changes found in function parameters")
+
+        updated_document = await self.service.update_document(
```
YS-L, line 147: Do we have to validate that the new parameters are actually valid for the function defined in the data warehouse?

featurebyte/routes/user_defined_function/controller.py
```diff
@@ -0,0 +1,226 @@
...
+
+        # check if function used in any saved feature
+        features = await self.feature_service.list_documents(
+            query_filter={"user_defined_function_ids": {"$in": [document_id]}},
+        )
+        if features["total"]:
+            feature_names = [doc["name"] for doc in features["data"]]
+            raise DocumentDeletionError(
+                f"User defined function used by saved feature(s): {feature_names}"
+            )
```
YS-L, line 178: This parts seems to be used more than once. Might be useful to refactor as a helper function.

featurebyte/routes/user_defined_function/controller.py
gee-senbong: The error message will be kept inside `{exc}`. Will update this part to drop `User defined function not exists: `.

featurebyte/routes/user_defined_function/controller.py
gee-senbong: I think you are right, will relax the check here.

featurebyte/routes/user_defined_function/controller.py
gee-senbong: Yes, we should validate it. Good catch.

YS-L (approved):
LGTM

PR #1470, opened 2023-07-04
feature: add user defined function SDK api object

gee-senbong:
## Description

This PR introduces user-defined-function SDK api object.

## Type of Change

## Checklist


featurebyte/api/user_defined_function.py
```diff
@@ -0,0 +1,319 @@
...
+        Associated feature store id of the user-defined function.
+
+        Returns
+        -------
+        PydanticObjectId
+        """
+        return self.cached_model.feature_store_id
+
+    @property
+    def function_name(self) -> str:
```
YS-L, line 154: Would it be clearer to name this `sql_function_name` to distinguish this from the user defined function's `name`?

featurebyte/api/user_defined_function_injector.py
```diff
@@ -0,0 +1,331 @@
...
+                name=None,
+                node_name=node.name,
+                feature_store=extracted_params.feature_store,
+                tabular_source=extracted_params.tabular_source,
+                dtype=udf.output_dtype,
+            )
+
+        # assign the dynamic method to the function accessor
+        assert udf.name is not None
+        setattr(func_accessor, udf.name, _method_wrapper)
```
YS-L, line 313: Is there already some validation that the udf name is a valid python identifier?

gee-senbong: Good catch. Current validation is only done the `function_name` but not `name`.

featurebyte/query_graph/node/function.py
```diff
@@ -223,6 +234,16 @@ def _derive_sdk_code(
...
+            node_name=self.name,
+        )
+        expression = get_object_class_from_function_call(
+            self.parameters.name,
+            *function_parameters,
+            module_path="featurebyte",
+            class_name="UDF",
+        )
+        statements: List[StatementT] = [(out_var_name, expression)]
+        return statements, out_var_name
```
gee-senbong, line 249: This is a temporary solution. We need to use an explicit method to construct UDF function
(https://featurebyte.atlassian.net/browse/DEV-1878).

Current approach will have issue if the following happens:
1. User create a global UDF and then create a feature using the global UDF.
2. User create a local UDF with the same name as the global one (the method in UDF accessor will be overridden by the local one).
3. User save the previously created feature (using global UDF). This will cause error as the feature definition execution will use the local UDF. The final generated graph will be different from the original graph.

featurebyte/api/user_defined_function_injector.py
```diff
@@ -0,0 +1,331 @@
...
+            check_type(
+                param.name, value=value, expected_type=get_param_type_annotations(param.dtype)
+            )
+
+            # prepare generic function parameter
+            func_param: FunctionParameterInput
+            if isinstance(value, (ViewColumn, Feature)):
+                if value.dtype != param.dtype:
+                    raise TypeError(
+                        f'Input ViewColumn or Feature "{value.name}" has dtype {value.dtype} '
```
YS-L, line 160: Will the `value.name` be empty if the column or feature is unnamed? In that case might be useful to mention the name of the violated argument in the error message.

featurebyte/api/user_defined_function.py
gee-senbong: Create a ticket to update this (https://featurebyte.atlassian.net/browse/DEV-1880) as it would be better to fix it at the model level too. Is it ok to you?

YS-L: Sure 👍

featurebyte/api/user_defined_function.py
```diff
@@ -0,0 +1,319 @@
...
+            udf = UserDefinedFunctionModel(**udf_dict)
+            assert udf.name is not None, "User-defined function name cannot be None."
+            if udf.name in name_to_udf_dict and udf.catalog_id is None:
+                # if udf is global but there is already a UDF with the same name (which must be local),
+                # skip the global UDF as it is overridden by the local one
+                continue
+            name_to_udf_dict[udf.name] = udf
+    except InvalidSettingsError:
+        # ignore invalid settings error due to fail to connect to the server
+        logger.info("Failed to synchronize user-defined functions.")
```
YS-L, line 76: Is it intentional that we don't raise an error that is visible to the user here?

YS-L (approved):
LGTM

featurebyte/api/user_defined_function.py
gee-senbong: Yes. This error will be thrown when we don't set the user profile at the beginning.

PR #1477, opened 2023-07-05
[DEV-1824] target: add preview, and integration test

jevonyeoh:
## Description
This PR makes a few additions to the `Target` API.

- adds a `preview` endpoint as a way to materialize some values of Target
- adds an end-to-end integration test
- have `Target` extend the `Series` class. Note that we'll still need to make some changes here to only allow `Targets` to interact with other `Target`s. Will also add more tests for this in a follow-up. Mostly added it here to unblock some of the code for the end-to-end test.

## Related Issue
https://featurebyte.atlassian.net/browse/DEV-1824

## Type of Change

## Checklist


Taskfile.yaml
```diff
@@ -311,6 +311,12 @@ tasks:
     cmds:
       - poetry run pytest --junitxml=pytest.xml -k playground tests/notebooks
 
+  test-reset:
```
jevonyeoh, line 314: a small utility which i used when trying to run integration tests

featurebyte/api/target.py
```diff
@@ -67,38 +100,85 @@ def horizon(self) -> Optional[str]:
...
+        -------
+        Node
+        """
+        try:
+            node_to_return = self.cached_model.node
+        except RecordRetrievalException:
+            node_name = self.internal_node_name
+            assert node_name is not None
+            node_to_return = self.graph.get_node_by_name(node_name)
+        return node_to_return
```
gee-senbong, line 118: Curious where this node is used? 
```
node_to_return = self.cached_model.node   <---- this is the node of the pruned graph (stored at the persistent)
```
If it is desired to access the node on the global graph, should use something like `global_graph.get_node_by_name(sef.node_name)`.

featurebyte/api/target.py
```diff
@@ -67,38 +100,85 @@ def horizon(self) -> Optional[str]:
...
+
+        Returns
+        -------
+        FeatureModel
+        """
+        pruned_graph, mapped_node = self.extract_pruned_graph_and_node()
+        target_dict = self.dict(by_alias=True)
+        target_dict["graph"] = pruned_graph.dict()
+        target_dict["node_name"] = mapped_node.name
+        return TargetModel(**target_dict)
```
gee-senbong, line 132: The `graph` & `node_name` of `self.dict(by_alias=True)` is already a pruned graph. Maybe we could drop the statement at line 128 (https://github.com/featurebyte/featurebyte/blob/cc24823e663e72e1a9369785d33950c637ce57cd/featurebyte/core/generic.py#L224-L236).

featurebyte/service/preview.py
```diff
@@ -436,6 +437,68 @@ async def preview_featurelist(
...
+            source_type=feature_store.type,
+            parent_serving_preparation=parent_serving_preparation,
+            skip_tile_generation=True,
+        )
+        result = await session.execute_query(preview_sql)
+        if result is None:
+            return {}
+        if updated:
+            result = result.drop(SpecialColumnName.POINT_IN_TIME, axis="columns")
+        return dataframe_to_json(result)
```
gee-senbong, line 500: Curious why don't we reuse the feature preview route? 🤔

jevonyeoh: ah possible now, i previously had a separate helper function that was being called from here, but now that that's cleaned up, i should be able to reuse the feature preview route here too i tihnk.

featurebyte/api/target.py
jevonyeoh: ah think this was sth that i added when tinkering around with a previous version, doesn't seem used now so i removed it!

Taskfile.yaml
YS-L: Curious under which situation will we need to run this?

featurebyte/query_graph/sql/preview.py
YS-L, line 1: Can the filename be `feature_target_preview.py`? There is already another `preview.py` file for non-feature preview: https://github.com/featurebyte/featurebyte/blob/main/featurebyte/query_graph/sql/interpreter/preview.py

I'm actually fine to keep this file as `feature_preview.py`. Target is really not that much different from feature in terms of sql generation.

featurebyte/api/aggregator/forward_aggregator.py
```diff
@@ -81,12 +83,17 @@ def forward_aggregate(
             input_nodes=[forward_aggregate_node],
         )
         # Build and return Target
+        agg_method = construct_agg_func(agg_func=cast(AggFunc, method))
+        output_var_type = self.get_output_var_type(agg_method, method, value_column)
```
YS-L, line 87: I thought the var type is already derived from graph / operation structure automatically. Is that true @gee-senbong?

gee-senbong: It is derived from the operation structure during feature saving (required to call `FeatureModel` constructor, https://github.com/featurebyte/featurebyte/blob/f024c4b2a894abcd103bc0337173f030d6bdb2e3/featurebyte/models/feature.py#L224-L252). Currently, it is not derived automatically for SDK API object.

featurebyte/query_graph/sql/aggregator/forward.py
```diff
@@ -63,6 +63,10 @@ def _get_aggregation_subquery(self, specs: list[ForwardAggregateSpec]) -> LeftJo
         horizon_in_seconds = 0
         if spec.parameters.horizon:
             horizon_in_seconds = pd.Timedelta(spec.parameters.horizon).total_seconds()
+        end_point_expr_current: Expression = cast(
+            Expression,
+            parse_one(f"FLOOR({point_in_time_epoch_expr.sql()})"),
```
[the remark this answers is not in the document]
jevonyeoh: ah this was copied from the feature sql, i'll remove the flooring here since it isn't needed for targets

featurebyte/query_graph/sql/preview.py
```diff
@@ -24,15 +24,17 @@
...
 
-def get_feature_preview_sql(
+def get_feature_or_target_preview_sql(
     request_table_name: str,
     graph: QueryGraphModel,
     nodes: list[Node],
     source_type: SourceType,
     point_in_time_and_serving_name_list: Optional[list[dict[str, Any]]] = None,
     parent_serving_preparation: Optional[ParentServingPreparation] = None,
+    skip_tile_generation: bool = False,
```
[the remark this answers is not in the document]
jevonyeoh: oops forgot to remove!

featurebyte/api/target.py
jevonyeoh: hmm, there seems to be some errors if i don't explicitly call the pruning. for what it's worth, the feature class also does this explicitly pruning. i'll keep this for now and we can look to remove later if that's ok?

featurebyte/query_graph/sql/preview.py
jevonyeoh: sounds good - renamed back to `feature_preview.py`

Taskfile.yaml
jevonyeoh: i removed it for now, but there were some situations where my db state seemed corrupted, so i was just running this to reset, and restart my local test env in one command

featurebyte/query_graph/sql/aggregator/forward.py
```diff
@@ -63,22 +63,26 @@ def _get_aggregation_subquery(self, specs: list[ForwardAggregateSpec]) -> LeftJo
         horizon_in_seconds = 0
         if spec.parameters.horizon:
             horizon_in_seconds = pd.Timedelta(spec.parameters.horizon).total_seconds()
+        end_point_expr_current: Expression = cast(
+            Expression,
+            parse_one(f"{point_in_time_epoch_expr.sql()}"),
```
YS-L, line 68: Usually there isn't a need for `parse_one()`. Here `parse_one(f"{point_in_time_epoch_expr.sql()}")` is equivalent to simply `point_in_time_epoch_expr`.

featurebyte/query_graph/sql/aggregator/forward.py
```diff
@@ -63,22 +63,26 @@ def _get_aggregation_subquery(self, specs: list[ForwardAggregateSpec]) -> LeftJo
...
+        )
         end_point_expr: Expression = cast(
             Expression,
-            parse_one(f"FLOOR({point_in_time_epoch_expr.sql()} + {horizon_in_seconds})"),
+            parse_one(f"{point_in_time_epoch_expr.sql()} + {horizon_in_seconds}"),
```
YS-L, line 72: This could use
```python
end_point_expr = expressions.Add(this=point_in_time_epoch_expr, expression=make_literal_value(horizon_in_seconds))
```

featurebyte/query_graph/sql/aggregator/forward.py
```diff
@@ -63,13 +64,9 @@ def _get_aggregation_subquery(self, specs: list[ForwardAggregateSpec]) -> LeftJo
...
         if spec.parameters.horizon:
             horizon_in_seconds = pd.Timedelta(spec.parameters.horizon).total_seconds()
-        end_point_expr_current: Expression = cast(
-            Expression,
-            parse_one(f"{point_in_time_epoch_expr.sql()}"),
-        )
-        end_point_expr: Expression = cast(
-            Expression,
-            parse_one(f"{point_in_time_epoch_expr.sql()} + {horizon_in_seconds}"),
+        end_point_expr_current: Expression = cast(Expression, point_in_time_epoch_expr)
```
YS-L, line 67: nit: I suspect the cast and explicit type annotation is unnecessary.

YS-L (approved):
LGTM

featurebyte/query_graph/sql/aggregator/forward.py
jevonyeoh: haha yes, forgot to rm.

featurebyte/api/target.py
gee-senbong: Since this is only used for previewing purpose, we can keep it now. 👍

gee-senbong (approved):

PR #1493, opened 2023-07-07
[chore] integration/conftest: use load dataframes from CSVs

jevonyeoh:
## Description
Use CSVs for loading data. It's faster to load a dataframe from a CSV, compared to generating them on the fly. I also personally found it easier to debug/trace through some logic when just looking at a raw CSV.

## Type of Change

## Checklist


gee-senbong (approved):
LGTM

jevonyeoh:
@gee-senbong - fyi, i updated a few more dataframes -> csvs.

gee-senbong:
> @gee-senbong - fyi, i updated a few more dataframes -> csvs.

Could we reduce the csv file?

gee-senbong (commented):
Would prefer to reduce the csv file to reduce the file size.

jevonyeoh:
oops forgot about this PR. hm i think all the rows in the CSV are used as part of the tests so i'm not sure if i can cut down the file size. just to understand, is the concern around the file size the readability of the file? or are there are concerns?

gee-senbong:
> oops forgot about this PR. hm i think all the rows in the CSV are used as part of the tests so i'm not sure if i can cut down the file size. just to understand, is the concern around the file size the readability of the file? or are there are concerns?

Usually it is not a good practice to commit huge fixture files to the repo. Curious how much improvement based on using generated fixtures (usually the pandas dataframe construction here should be pretty fast).

gee-senbong:
If the fixture file is huge, everything we update the fixture generation logic will cause this to be committed to git history. This will make the current repository size growing bigger & bigger. Would be good to consider the tradeoff here.

@YS-L What do you think?

jevonyeoh:
for construction vs. reading, it's about 15x faster, but they're pretty small times so not super significant (eg. 0.27s, vs 0.015s)

the few benefits i had in mind was:
- easier to read a CSV, compared to reading through the logic used to generate a dataframe. though, after thinking about it, if the file is that big, it's not that useful anyway.
- can probably clean up some of the test code to not hold these dataframes as session scoped, since we should just be able to read them once, and just load them into the warehouse.

with that in mind, i too am not too huge a fan of having large CSVs in the repo. i'll just leave it as is for now.

PR #1508, opened 2023-07-11
Add target definition support

gee-senbong:
## Type of Change

## Checklist


featurebyte/api/target.py
```diff
@@ -108,6 +114,21 @@ def horizon(self) -> Optional[str]:
...
+            example=textwrap.dedent(
+                """
+                >>> target = catalog.get_target("InvoiceCount_60days")  # doctest: +SKIP
+                >>> target_definition = target.definition  # doctest: +SKIP
+                """
+            ).strip(),
+        )
+    )
+    def definition(self) -> str:  # pylint: disable=missing-function-docstring
+        return self._generate_definition()
```
gee-senbong, line 130: @YS-L @jevonyeoh 

Would like to get the early feedback of this pattern.

featurebyte/api/feature_or_target_mixin.py
```diff
@@ -0,0 +1,73 @@
...
+then transpiled into platform-specific SQL (e.g. SnowSQL, SparkSQL) for {object_type} materialization.
+
+Returns
+-------
+str
+
+Examples
+--------
+{example}
+"""
```
gee-senbong, line 33: This is common docstring template shared by feature & target (maybe should move this into a separate module).

featurebyte/api/feature_or_target_mixin.py
```diff
@@ -0,0 +1,73 @@
...
+    ----------
+    new_docstring: str
+        New docstring to use
+
+    Returns
+    -------
+    Func
+    """
+
+    def decorator(func: Func) -> Func:
```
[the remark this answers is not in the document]
gee-senbong: Tested it locally, seems like it will void the type checking. Since this is used mostly at the SDK api object public method, the impact should be less severe than those internal methods. If this is a serious concern and there is no other way to get around this, we may need to look for other approach to share the doc string.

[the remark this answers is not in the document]
gee-senbong: I created a test method at `Feature` object:
```
    def test(self) -> int:
        return self.definition
```

I get the following error when running mypy:
Screenshot 2023-07-11 at 3 55 13 PM

Seems like it still can detect the error.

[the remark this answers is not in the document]
YS-L: I see, seems like it's a non-issue 👍

featurebyte/api/target.py
gee-senbong: Ya, thinking of making the decorator as a class too to make it handle those indentation issue. If we follow this pattern, it should helps to reduce the code lines of these api objects. Personally, i think it is more readable (focus more on the code & the differences).

jevonyeoh: I think I would also lean towards repeating docstrings for simplicity. Though, I do like that in your approach, we can have more consistent docstrings. 🤔 

Separately, some other tools might not pick up programatically generated docs. For example, in the IDE itself, mouseover docs won't have the full docstring.

Screenshot 2023-07-11 at 4 12 16 PM

gee-senbong: Ya, seems like pycharm does not show the docstring. Tested jupyter notebook just now, it can render the docstring properly.

gee-senbong: The main reason of doing this is because i found there are some discrepancies between similar methods in `Feature` and `Target` (`primary_entity` in feature vs `entities` in target). If we start doing it now, it will force us to revise only single docstring. In addition, we may also need to check the consistencies between two classes.

YS-L (approved):

featurebyte/api/target.py
gee-senbong: @jevonyeoh @YS-L 

The decorator is revised here to make it clearer (like having different sections string).

PR #1514, opened 2023-07-12
[DEV-1935] api/target_namespace: add API object, and add create function

jevonyeoh:
## Description
This PR adds the `TargetNamespace` API object, and also allows us to create a `TargetNamespace` object to register spec-like versions of a Target that don't necessarily have a recipe.

## Related Issue
https://featurebyte.atlassian.net/browse/DEV-1935

## Type of Change

## Checklist


featurebyte/schema/target_namespace.py
```diff
@@ -24,11 +24,11 @@ class TargetNamespaceCreate(FeatureByteBaseModel):
 
     id: Optional[PydanticObjectId] = Field(default_factory=ObjectId, alias="_id")
     name: StrictStr
-    dtype: DBVarType
+    dtype: DBVarType = Field(default=DBVarType.UNKNOWN)
```
gee-senbong, line 27: Should we make this `Optional[DBVarType]`?

featurebyte/routes/target_namespace/api.py
```diff
@@ -29,6 +32,16 @@
 router = APIRouter(prefix="/target_namespace")
 
 
+@router.post("", response_model=TargetNamespaceModel, status_code=HTTPStatus.CREATED)
+async def create_target(request: Request, data: TargetNamespaceCreate) -> TargetNamespaceModel:
```
gee-senbong, line 36: nit: `create_target_namespace`

featurebyte/schema/target_namespace.py
gee-senbong: Or `dtype` should be a required parameter during target namespace creation

jevonyeoh: hm i think there's no requirement for the dtype to be known during namespace creation at the moment. will update to optional!

gee-senbong: If there is no requirement, we should use `Optional`. Thinking whether we should check with Xavier whether to include this.

featurebyte/api/target_namespace.py
```diff
@@ -0,0 +1,145 @@
...
+        Returns
+        -------
+        PydanticObjectId
+        """
+        return self.cached_model.default_target_id
+
+    @property
+    def default_version_mode(self) -> DefaultVersionMode:
+        """
+        Default feature namespace version mode of this feature namespace
```
gee-senbong, line 130: Should be target namespace.

featurebyte/api/target_namespace.py
```diff
@@ -0,0 +1,145 @@
...
+        Returns
+        -------
+        List[PydanticObjectId]
+        """
+        return self.cached_model.target_ids
+
+    @property
+    def entity_ids(self) -> List[PydanticObjectId]:
+        """
+        List of entity IDs used by the feature namespace
```
gee-senbong, line 105: Should be target namespace

featurebyte/api/target_namespace.py
```diff
@@ -0,0 +1,145 @@
...
+from featurebyte.api.entity import Entity
+from featurebyte.api.savable_api_object import SavableApiObject
+from featurebyte.exception import RecordRetrievalException
+from featurebyte.models.base import PydanticObjectId
+from featurebyte.models.feature_namespace import DefaultVersionMode
+from featurebyte.models.target_namespace import TargetNamespaceModel
+from featurebyte.schema.target_namespace import TargetNamespaceUpdate
+
+
+class TargetNamespace(SavableApiObject):
```
gee-senbong, line 23: Should we refactor this & `FeatureNamespace`? Since both of them share many similarities.

featurebyte/api/base_feature_target_namespace.py
```diff
@@ -0,0 +1,46 @@
+"""
```
gee-senbong, line 1: Should we make the naming convention consistent with `feature_or_target_mixin.py`? If current naming convention is preferred, we could update `feature_or_target_mixin.py` to following single convention.

featurebyte/api/base_feature_target_namespace.py
```diff
@@ -0,0 +1,46 @@
...
+
+from pydantic import Field
+
+from featurebyte.api.api_object import ApiObject
+from featurebyte.exception import RecordRetrievalException
+from featurebyte.models.base import PydanticObjectId
+from featurebyte.models.feature_namespace import BaseFeatureNamespaceModel, DefaultVersionMode
+
+
+class BaseFeatureTargetNamespace(ApiObject):
```
gee-senbong, line 14: Should we make it a mixin rather than a base class (similar to `feature_or_target_mixin.py`)?

jevonyeoh: sure i can rename it, but just to clarify though, a mixin and a base class function pretty much the same way yea?

gee-senbong: From the code wise, both rely on python inheritance. There are some minor differences (https://stackoverflow.com/questions/860245/mixin-vs-inheritance):
Base class => Usually object is IS inheritance check
Mixin => mixin is rarely useful as a standalone object (it is like introducing special properties to the class)

gee-senbong: In this case, it is more like documentation purpose. If a class has name `Mixin`, we don't expect it to be instantiated and it should be inherited by some other classes.

tests/fixtures/request_payloads/target_namespace.json
```diff
@@ -0,0 +1,19 @@
+{
+    "_COMMENT": [
+        "THIS IS A GENERATED FILE. DO NOT EDIT THIS FILE DIRECTLY.",
+        "Instead, update the test tests/unit/test_generate_payload_fixtures.py#test_save_payload_fixtures.",
+        "Run `pytest --update-fixtures` to update it."
+    ],
+    "_id": "64ae4be43a93459ede8c383b",
+    "default_target_id": "64a80107d667dd0c2b13d8cd",
+    "default_version_mode": "AUTO",
+    "dtype": "UNKNOWN",
```
gee-senbong, line 10: Is this fixture get updated? Just to check whether we have remove `UNKNOWN` from default `dtype`.

gee-senbong (approved):
LGTM 👍

tests/fixtures/request_payloads/target_namespace.json
jevonyeoh: good catch - updated.

PR #1524, opened 2023-07-14
[chore] Update `extract_operation_structure`'s `keep_all_source_columns` parameter

gee-senbong:
## Description

Update `OperationStructureExtractor`'s `keep_all_source_columns` parameter handling so that it is `True` for all cases except when using to derive feature/target info's `metadata` field.

## Related Issue

The rationale of this change is due to https://featurebyte.atlassian.net/browse/DEV-1943. Currently, `keep_all_source_columns` is `False` by default. This is very risky as it may drops those columns that contributes only to the changes of the row index lineage. To prevent similar issue happens in the future, this PR makes `keep_all_source_columns` to be `True` for most cases.

## Type of Change

## Checklist

### DIFF
--- a/featurebyte/query_graph/graph.py
+++ b/featurebyte/query_graph/graph.py
@@ -280,7 +280,12 @@ class QueryGraph(QueryGraphModel):
             node_name_map[node.name] = node_global.name
         return self, node_name_map
 
-    def extract_operation_structure(self, node: Node, **kwargs: Any) -> OperationStructure:
+    def extract_operation_structure(
+        self,
+        node: Node,
+        keep_all_source_columns: bool,
+        **kwargs: Any,
+    ) -> OperationStructure:
         """
         Extract operation structure from the graph given target node
 
@@ -288,6 +293,8 @@ class QueryGraph(QueryGraphModel):
         ----------
         node: Node
             Target node used to construct the operation structure
+        keep_all_source_columns: bool
+            Whether to keep all source columns in the operation structure
         kwargs: Any
             Additional arguments to be passed to the OperationStructureExtractor.extract() method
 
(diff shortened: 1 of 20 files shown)


featurebyte/query_graph/graph.py
```diff
@@ -280,22 +280,31 @@ def load(self, graph: QueryGraphModel) -> Tuple["QueryGraph", Dict[str, str]]:
             node_name_map[node.name] = node_global.name
         return self, node_name_map
 
-    def extract_operation_structure(self, node: Node, **kwargs: Any) -> OperationStructure:
+    def extract_operation_structure(
+        self,
+        node: Node,
+        keep_all_source_columns: bool,
```
jevonyeoh, line 286: Would it make sense/be possible to set the default value here to be True?

gee-senbong: Yes, that's my initial fix. In the end, feel like making it explicit is better so that it forces us to be aware this parameter. I don't have strong opinion here.

YS-L: If there are only limited cases where it should be set as False and it's risky to do so, having this default to True makes sense to me.

YS-L (approved):
Curious if this fixes the pruning issue in https://featurebyte.atlassian.net/browse/DEV-1943?

gee-senbong:
> https://featurebyte.atlassian.net/browse/DEV-1943

Nope, there will be another PR to fix this. This PR only fixes one of the root cause.

featurebyte/query_graph/graph.py
gee-senbong: Will update this in the coming fix.

PR #1546, opened 2023-07-19
[DEV-1955] Support description update and retrieval for saved objects

kchua78:
## Description

Support description update and retrieval for saved objects

## Related Issue

https://featurebyte.atlassian.net/browse/DEV-1955

## Type of Change

## Checklist

### DIFF
--- a/featurebyte/api/api_object.py
+++ b/featurebyte/api/api_object.py
@@ -401,6 +401,8 @@ class ApiObject(FeatureByteBaseDocumentModel, AsyncMixin):
         update_payload: Dict[str, Any],
         allow_update_local: bool,
         add_internal_prefix: bool = False,
+        url: Optional[str] = None,
+        skip_update_schema_check: bool = False,
     ) -> None:
         """
         Update object in the persistent
@@ -415,6 +417,10 @@ class ApiObject(FeatureByteBaseDocumentModel, AsyncMixin):
             Whether to add internal prefix (`internal_`) to the update key (used when the attribute to be updated
             starts with `internal_`). This flag only affects local update behavior (no effect if `allow_update_local`
             is disabled).
+        url: Optional[str]
+            URL to update the object
+        skip_update_schema_check: bool
+            Whether to skip update schema check
 
         Raises
         ------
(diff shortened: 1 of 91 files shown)


featurebyte/api/api_object.py
```diff
@@ -544,3 +544,34 @@ def info(self, verbose: bool = False) -> Dict[str, Any]:
...
+            self._update_cache(object_dict)  # update object cache
+            type(self).__init__(
+                self,
+                **object_dict,
+                **self._get_init_params_from_object(),
+            )
+        elif response.status_code == HTTPStatus.NOT_FOUND:
+            self.description = description
+        else:
+            raise RecordUpdateException(response=response)
```
gee-senbong, line 577: Would it better to update `def update` method to accept `url` parameter, and then call it by:
```
self.update(
    update_payload={"description": description},
    allow_update_local=False,
    url=f"{self._route}/{self.id}/description",
)
```

Doing this, we could avoid setting description before the object is saved (`allow_update_local=False`) and also reuse update logic (like update cached handling, exception handling etc).

gee-senbong: We may also need to add a `description` property here to retrieved the description value. Something like
```
@property
def description(self):
        return self.cached_model.description:
```

featurebyte/routes/batch_feature_table/api.py
```diff
@@ -146,3 +147,20 @@ async def download_table_as_pyarrow_table(
         get_credential=request.state.get_credential,
     )
     return result
+
+
+@router.patch("/{batch_feature_table_id}/description", response_model=BatchFeatureTableModel)
+async def update_batch_feature_table_description(
```
jevonyeoh, line 153: seems like we mostly duplicate this logic across all the routers - i wonder if it's possible to reduce boilerplate by putting router paths into a base class, and having child routers inherit from that.

maybe sth like https://stackoverflow.com/questions/63853813/how-to-create-routes-with-fastapi-within-a-class?

featurebyte/service/base_document.py
```diff
@@ -972,3 +973,35 @@ async def remove_block_modification_by(
...
+        update_dict = {"description": description}
+        updated_document = self.document_class(**{**document.dict(by_alias=True), **update_dict})
+        if document != updated_document:
+            # only update if change is detected
+            await self.persistent.update_one(
+                collection_name=self.collection_name,
+                query_filter=self._construct_get_query_filter(document_id=document_id),
+                update={"$set": update_dict},
+                user_id=self.user.id,
+            )
```
gee-senbong, line 1007: Would be good if we could refactor the common logic between `update_documents` and `update_document_description` so that it is easier to ensure consistency between these two update methods.

featurebyte/routes/batch_feature_table/api.py
kchua78: Great idea.  Will look into this in a separate PR since this one is rather large.

gee-senbong (approved):
LGTM

PR #1553, opened 2023-07-21
[DEV-1969] Configurations refactoring to singleton

bushwhackr:
## Description

Motivation for Singleton, given the below yaml.

```yaml
# ~/.featurebyte/config
profiles:
  - name: test
    ...
---
# config2.yaml
profiles:
  - name: test2
    ...
```

```python
config = Configurations()
config.use_profile("test")  # Okay

config2 = Configurations("config2.yaml")
config2.use_profile("test2")  # Error, its looking for profile in Configuration().profiles
print(config2.profiles[0].name)  # "test2"  
^^^^^^^^^^^^^^^^^^^^^^^ its weird that you can see a "test2" profile inside config2 but unable to use_profile() it
```

+ Migration to singleton Configuration()
+ Moving use_profile to a singleton
+ Upgraded mypy + added more required types package
+ Small change in how changelog entries are rendered in PR

This is a very specific use case that most users will not encounter. As most users just use the default `import featurebyte as fb; fb.use_profile()` it will use the default profile from `${FEATUREBYTE_HOME}/config.yaml`

## Type of Change

## Checklist


featurebyte/config.py
```diff
@@ -271,20 +271,58 @@ class Configurations:
...
+            self.profiles: List[Profile] = []
+            self.settings: Dict[str, Any] = {}
+            self.logging: LoggingSettings = LoggingSettings()
+            self.reload(config_file_path)
+
+        # Force specified or config file path specified
+        elif force or config_file_path is not None:
+            self.reload(config_file_path)
+        else:
+            pass  # do nothing
```
bushwhackr, line 307: singleton impl

featurebyte/config.py
```diff
@@ -357,25 +390,28 @@ def _parse_config(self, path: Path) -> None:
...
             default_profile = self.settings.pop("default_profile", None)
-            selected_profile_name = os.environ.get("FEATUREBYTE_PROFILE")
-            if selected_profile_name:
-                self._profile = profile_map.get(selected_profile_name)
-            else:
-                self._profile = profile_map.get(default_profile)
+            self._profile = profile_map.get(default_profile)
+        else:
+            self.profiles = []
+            self._profile = None
```
bushwhackr, line 414: due to the new singleton, if config files are missing fields, have to force default values back, the getters will return the oldest valid value, which might not be present in the "active" config file

featurebyte/config.py
```diff
@@ -434,21 +469,22 @@ def use_profile(cls, profile_name: str) -> None:
...
-            cls.check_sdk_versions()
+            self._profile = new_profile
+            self.check_sdk_versions()
         except InvalidSettingsError:
             # restore previous profile
-            if current_profile_name:
-                os.environ["FEATUREBYTE_PROFILE"] = current_profile_name
-            else:
-                os.environ.pop("FEATUREBYTE_PROFILE", None)
+            self._profile = old_profile
```
bushwhackr, line 487: No need to maintain `FEATUREBYTE_PROFILE` env anymore. Every Configurations().profile would have a singluar value.

featurebyte/config.py
```diff
@@ -271,20 +271,58 @@ class Configurations:
         skipped_members=["get_client"],
     )
 
-    def __init__(self, config_file_path: Optional[str] = None) -> None:
+    _instance = None
+
+    def __new__(cls, *args: Any, **kwargs: Any) -> "Configurations":
+        if Configurations._instance is None:
```
jevonyeoh, line 277: there's a singleton meta class that could potentially be used i think?

https://github.com/featurebyte/featurebyte/blob/main/featurebyte/common/singleton.py

jevonyeoh (commented):
what's the motivation for making this singleton?

featurebyte/config.py
bushwhackr: I just checked. It seems like the flow when not init is okay. But when it is initialized, it hits the `SingletonMeta::__call__` and ends there.
```python
def __call__(cls, *args: Any, **kwargs: Any) -> Any:
    if cls not in cls._instances:
        instance = super().__call__(*args, **kwargs)
        cls._instances[cls] = instance
    return cls._instances[cls]
```

However, the early return will mean that I am unable to reinitalize a different config file if a user specifies another config file to use.

bushwhackr: That is to say, 
```python
# This is init-ed to ~/.featurebyte/config.yaml
config = Configuration()
# /asd.yaml will never be read because the SingletonMeta will just return the already initialized instance
config2 = Configuration("/asd.yaml")  
```

bushwhackr: for motivation, see description edit

featurebyte/config.py
jevonyeoh: could we reset the instance variables in the constructor as well, i.e reset them under the `force or config_file_path present` path? that way i think we should be able to skip all the `else` blocks here

jevonyeoh (commented):
do we have any tests in `test_config` that covers the singleton logic?

.changelog/DEV-1969.yaml
```diff
@@ -0,0 +1,38 @@
...
+
+# SAMPLE
+#subtext: |
+#  + note this will make everything better
+#  + major performance improvement
+#  + time reduction for test suite from 10 minutes to 2 minutes
+#  ```
+#   Code sample
+#   goes here
+#  ```
```
gee-senbong, line 38: Is it intentional to have 2 change log entries?

bushwhackr: I'll just drop the singleton enhancement one, since it doesn't affect UX.

featurebyte/worker/task_executor.py
```diff
@@ -130,6 +131,12 @@ def _setup_worker_config(self) -> None:
             "default_profile: worker\n\n",
             encoding="utf-8",
         )
+        # NOTE: Unable to remove this try-except block
+        # due to test_task_manager expecting a server listening on 127.0.0.1:8080/status endpoint
```
YS-L, line 135: Would it be possible to patch test_task_manager to avoid this issue? Seems a bit risky to mask an exception here when that is not expected in actual code paths.

featurebyte/config.py
bushwhackr: refactored

featurebyte/config.py
```diff
@@ -271,37 +271,53 @@ class Configurations:
...
-            self._config_file_path.parent.mkdir(parents=True, exist_ok=True)
-            self._config_file_path.write_text(
-                "# featurebyte configuration\n\n"
-                "profile:\n"
-                "  - name: local\n"
-                "    api_url: http://127.0.0.1:8088\n\n"
-                "default_profile: local\n\n"
+            this will force a reload of configuration if specified
+        force: bool
+            Force reload configurations
```
YS-L, line 291: Would it be a bit clearer to name this parameter as `reload=True`?

featurebyte/worker/task_executor.py
```diff
@@ -130,6 +130,8 @@ def _setup_worker_config(self) -> None:
             "default_profile: worker\n\n",
             encoding="utf-8",
         )
+        # Reload newly written configuration
+        Configurations(force=True)
```
YS-L, line 134: I realise one user facing difference compared to previous behaviour is that within a python interpreter session, any changes to the config file will not have effect until `Configurations(force=True)` is run, whereas before the changes will be picked up immediately. Is that right @bushwhackr? 

The previous behaviour seems a bit more convenient, wondering if it's worth preserving (something like only caching `_config_file_path`, but always re-parse the config file when instantiating `Configurations`).

featurebyte/config.py
bushwhackr: Configurations("/config.yaml" , reload=False) will still reload. I think that might be mote confusing

featurebyte/worker/task_executor.py
bushwhackr: Okay if we always reparse I think there is no need to put the force/reload param

gee-senbong (commented):
Tested it locally, seems like some tests are taking my local configuration file. Would be good to fix this or create another ticket to track this issue.

featurebyte/worker/task_executor.py
bushwhackr: reverted reparsing code, due to lack of ability to run 2 servers at the same time.

bushwhackr:
> Tested it locally, seems like some tests are taking my local configuration file. Would be good to fix this or create another ticket to track this issue.

autouse=True is used for config file now so tests should not affect the users ~/.featurebyte/config.yaml

PR #1574, opened 2023-07-26
[DEV-1996] update compute_target to return observation table instead

jevonyeoh:
## Description
This PR updates the `compute_target_table` function to return an `ObservationTable` instead of a `TargetTable` object. We also make a few changes to start removing the `TargetTable` API object since that isn't needed anymore. 

This change aims to:
- make the SDK a bit simpler, and not introduce another API object
- allow us to reuse the `ObservationTable`, and be able to pass materialized Targets into the computation of historical features. This will allow us to get a complete set of training data.

Will follow-up to clean up target table a bit more, but getting this out first to get the SDK changes in so that we can release it.

## Related Issue
https://featurebyte.atlassian.net/browse/DEV-1996

## Type of Change

## Checklist


featurebyte/models/observation_table.py
```diff
@@ -28,8 +34,38 @@ class SourceTableObservationInput(SourceTableRequestInput):
...
+    type: Literal[RequestInputType.OBSERVATION_TABLE, RequestInputType.DATAFRAME]
+
+    async def materialize(
+        self,
+        session: BaseSession,
+        destination: TableDetails,
+        sample_rows: Optional[int],
+    ) -> None:
+        """
+        No-op materialize.
```
YS-L, line 53: Would be good to add a comment on why this is a no-op. Does this function actually get called for `TargetInput`?

jevonyeoh: good point - will add a comment and have it throw an exception, it shouldn't get called, but the linter throws an error since it expects it to be there!

YS-L (approved):

tests/integration/api/test_target_correctness.py
```diff
@@ -206,3 +207,11 @@ def test_forward_aggregate(
             expected_values["POINT_IN_TIME"], utc=True
         ).dt.tz_localize(None)
         fb_assert_frame_equal(df_targets, expected_values)
+
+        # Build full materialized Target observation table
```
YS-L, line 211: Would be nice if there is a test to cover the intended workflow: materialise a target, then use the resulting observation table to materialise some features.

jevonyeoh: i'll add in a follow-up!

PR #1581, opened 2023-07-27
[DEV-1970] Make deployment create/update a lock task

gee-senbong:
## Description

This PR 
* implements a `BaseLockTask` which can be used to run the task requires locking mechanism;
* makes deployment creation/update task a `BaseLockTask`.

## Type of Change

## Checklist


featurebyte/worker/task/base.py
```diff
@@ -99,3 +103,67 @@ async def execute(self) -> Any:
         """
         Execute the task
         """
+
+
+class BaseLockTask(BaseTask):
```
kchua78, line 108: 👍

kchua78 (approved):

featurebyte/worker/task/deployment_create_update.py
```diff
@@ -5,23 +5,43 @@
...
+
+    @property
+    def lock_blocking(self) -> bool:
+        # if the deployment is being created or updated by another task,
+        # fail the current task immediately without waiting
+        return False
+
+    def handle_lock_not_acquired(self) -> None:
+        payload = cast(DeploymentCreateUpdateTaskPayload, self.payload)
+        error_msg = f"Deployment {payload.output_document_id} is currently being created or updated"
```
YS-L, line 39: nit: Maybe can customise the error message based on deployment payload type.

featurebyte/worker/task/base.py
```diff
@@ -99,3 +103,67 @@ async def execute(self) -> Any:
...
+
+    async def execute(self) -> Any:
+        """
+        Execute the task
+
+        Returns
+        -------
+        Any
+        """
+        lock = self.app_container.redis.lock(self.lock_key)
```
YS-L, line 122: Do we have to configure a lock timeout to handle unexpected scenarios where the lock is not released by abruptly terminated tasks?

gee-senbong: Any suggestion for the lock expired time? If we add the lock at the task level, it is a bit difficult to set the expire value. That's why i did not include it in the current implementation.

featurebyte/worker/task/deployment_create_update.py
gee-senbong: We can't customize the error message based on the deployment payload type here unless we make another mongo call to retrieve the task payload (current type may not be the same as the one running on the other task).

featurebyte/worker/task/base.py
YS-L: Maybe we can set a large value like 24h as a default. The appropriate timeout could also be task dependent.

YS-L (approved):
LGTM

PR #1582, opened 2023-07-27
[DEV-1849] Support conditionally updating a feature using a mask derived from other feature(s)

YS-L:
## Description

The `row_index_lineage` alignment check is applicable to Columns but not Features. Features are joined into a common table and can freely interact with each other regardless of the lineage.

This overrides the `row_index_lineage` property for Feature class as a way to bypass the alignment checks. By doing this, we can support things like conditionally updating a feature using a mask derived from other features.

## Type of Change

## Checklist

### DIFF
--- a/featurebyte/api/feature.py
+++ b/featurebyte/api/feature.py
@@ -130,6 +130,10 @@ class Feature(
 
     @property
     def row_index_lineage(self) -> Tuple[str, ...]:
+        # Override the row_index_lineage property to return a constant value. This is so that the
+        # check for row_lineage_index alignment for Feature objects always passes. The check is not
+        # applicable to Feature objects because they can freely interact with each other regardless
+        # of their lineage.
         return tuple()
 
     @root_validator(pre=True)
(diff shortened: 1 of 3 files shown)


featurebyte/api/feature.py
```diff
@@ -128,6 +128,10 @@ def _get_feature_tiles_specs(self) -> List[Tuple[str, List[TileSpec]]]:
         tile_specs = ExtendedFeatureModel(**self.dict(by_alias=True)).tile_specs
         return [(str(self.name), tile_specs)] if tile_specs else []
 
+    @property
+    def row_index_lineage(self) -> Tuple[str, ...]:
+        return tuple()
```
gee-senbong, line 133: Would be good to leave some comment why it is empty. Is it because we did not use this value directly? Curious why don't we generate it from the operation structure.

gee-senbong (approved):

featurebyte/api/feature.py
YS-L: Sure, will add some comments. The intention is to make row index lineage check always pass for Feature objects, so the return value here can be any value as long as it's a constant.

I think there are a few possible approaches:

1. Update all places where we check whether `row_index_lineage` matches to ignore Feature 
2. Update derivation of `row_index_lineage` in operation structure to return a constant for Feature
3. Override the `row_index_lineage` for Feature class to return a constant (current approach)

Since 1 and 2 likely require more changes in multiple places, I opted the current approach.

PR #1607, opened 2023-08-02
[DEV-2026] Filter EventTable by tile date range directly when possible

YS-L:
## Description

This adds a filter on EventTable directly based on tile date range when possible. This is safe to do when there is no lag operations in the applied transforms. This mainly benefits the cases where a feature involves complex joins and the existing top level filter cannot be pushed down automatically. 

In testing, the speedup for the tile building query ranges from 1x (simple features based on EventTable only that won't benefit from this change) to more than 6x (features involving Event-Item-SCD joins + double aggregation; 2m 14s before vs 21s after).

## Type of Change

## Checklist

### DIFF
--- a/featurebyte/query_graph/sql/interpreter/tile.py
+++ b/featurebyte/query_graph/sql/interpreter/tile.py
@@ -3,19 +3,23 @@ This module contains the Query Graph Interpreter
 """
 from __future__ import annotations
 
-from typing import cast
+from typing import Optional, cast
 
 from dataclasses import dataclass
 
 from featurebyte.enum import SourceType
+from featurebyte.query_graph.algorithm import dfs_traversal
 from featurebyte.query_graph.enum import NodeType
 from featurebyte.query_graph.model.graph import QueryGraphModel
 from featurebyte.query_graph.node import Node
 from featurebyte.query_graph.node.generic import GroupByNode
+from featurebyte.query_graph.sql.ast.base import EventTableTimestampFilter
 from featurebyte.query_graph.sql.builder import SQLOperationGraph
 from featurebyte.query_graph.sql.common import SQLType
 from featurebyte.query_graph.sql.interpreter.base import BaseGraphInterpreter
 from featurebyte.query_graph.sql.template import SqlExpressionTemplate
+from featurebyte.query_graph.transform.operation_structure import OperationStructureExtractor
+from featurebyte.query_graph.transform.pruning import prune_query_graph
 
 
 @dataclass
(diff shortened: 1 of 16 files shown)


tests/fixtures/expected_tile_sql_complex_feature_push_down_eligible.sql
```diff
@@ -0,0 +1,215 @@
...
+                    "col_char" AS "col_char",
+                    "col_text" AS "col_text",
+                    "col_binary" AS "col_binary",
+                    "col_boolean" AS "col_boolean",
+                    "event_timestamp" AS "event_timestamp",
+                    "cust_id" AS "cust_id"
+                  FROM "sf_database"."sf_schema"."sf_table"
+                  WHERE
+                    "event_timestamp" >= CAST(__FB_START_DATE AS TIMESTAMPNTZ)
+                    AND "event_timestamp" < CAST(__FB_END_DATE AS TIMESTAMPNTZ)
```
YS-L, line 113: This filter is now applied at a lower level here when selecting from event table.

featurebyte/query_graph/sql/interpreter/tile.py
```diff
@@ -99,25 +104,69 @@ def construct_tile_gen_sql(self, starting_node: Node) -> list[TileGenSql]:
...
+            groupby_input_node
+        )
+        op_struct = op_struct_info.operation_structure_map[groupby_input_node.name]
+        for column in op_struct.columns:
+            if (
+                column.name == groupby_node.parameters.timestamp
+                and isinstance(column, SourceDataColumn)
+                and column.table_id is not None
+            ):
+                assert isinstance(column, SourceDataColumn)
```
jevonyeoh, line 149: is this assert necessary since it's already in the if-statement?

jevonyeoh (commented):
is this only applicable for event tables, and not scd tables?

featurebyte/query_graph/sql/interpreter/tile.py
YS-L: Yes, this is necessary to make mypy happy. I think not necessary when the if condition is simpler.

YS-L:
> is this only applicable for event tables, and not scd tables?

Yes, this is only applicable for even tables. ChangeView / SCDTable features also use tiles, but they involve lag operations so the date filter cannot be pushed down. That is covered by this existing test case (no change to the sql there): https://github.com/featurebyte/featurebyte/blob/a9901c9dcfe9254f953322d92a85b8d5d9fb0f8c/tests/unit/api/test_change_view.py#L396-L467

jevonyeoh (approved):
and just for my understanding, why can't this approach be applied for those views with lag?

code changes lgtm

YS-L:
> why can't this approach be applied for those views with lag?

The reason is that `WHERE` clause is applied before the expressions (including lags) are evaluated within the select statement. If we apply the filter in those cases, the lag of the earliest row within the filter time range will be always be NA which is incorrect.

PR #1626, opened 2023-08-07
[DEV-2026] Apply timestamp filter for EventTable at lower level for ItemView

YS-L:
## Description

This updates the timestamp filtering logic to apply the event timestamp filter for EventTable at a lower level for ItemView.

## Type of Change

## Checklist

### DIFF
--- a/featurebyte/query_graph/sql/interpreter/tile.py
+++ b/featurebyte/query_graph/sql/interpreter/tile.py
@@ -159,6 +159,7 @@ class TileSQLGenerator:
                     if (
                         source_column.table_type == TableDataType.EVENT_TABLE
                         and source_column.dtype in {DBVarType.TIMESTAMP, DBVarType.TIMESTAMP_TZ}
+                        and source_column.table_id is not None
                     ):
                         return EventTableTimestampFilter(
                             timestamp_column_name=source_column.name,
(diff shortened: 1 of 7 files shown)


featurebyte/query_graph/sql/interpreter/tile.py
```diff
@@ -140,17 +140,31 @@ def _get_event_table_timestamp_filter(
...
+            # Check if the timestamp column is from an EventTable
+            if isinstance(column, SourceDataColumn):
+                if column.table_type == TableDataType.EVENT_TABLE and column.table_id is not None:
+                    return EventTableTimestampFilter(
+                        timestamp_column_name=column.name,
+                        event_table_id=column.table_id,
+                    )
+            elif isinstance(column, DerivedDataColumn) and set(column.transforms) == {"join"}:
+                # In ItemView, the event timestamp column is considered a derived column because
+                # it is derived from join.
```
gee-senbong, line 157: Any issue if we drop ` and set(column.transforms) == {"join"}`?
The reason that I am asking this question is that `column.transforms` attribute currently is not really used in any critical operation (only used in deriving feature info). Would be good if we could avoid using this attribute.

gee-senbong (approved):

featurebyte/query_graph/sql/interpreter/tile.py
gee-senbong: 🤔 I think i know why it is used here.

Other option is doing the following:
1. Extract the node from the graph based on `column.node_name`.
2. If the node parameter is from the `ItemViewGraphNodeParameters`, we are safe to apply the following filter.

PR #1664, opened 2023-08-21
[DEV-2093] vector_aggregation: support simple aggregates, and int values in vectors

jevonyeoh:
## Description
This PR makes a few updates to add support for
- simple aggregates
- aggregate_asat
- int values in vectors
- ensure we convert spark `ARRAY_TYPES` to lists properly

We do this by simplifying some of the UDAF logic, by assuming that we'll always return a list of doubles for all operations regardless of input vector type. This allows us to simplify some of the logic in the UDAF, and reduce the need to wrangle the types of the input in java.

## Related Issue
https://featurebyte.atlassian.net/browse/DEV-2075
https://featurebyte.atlassian.net/browse/DEV-2092
https://featurebyte.atlassian.net/browse/DEV-2093
https://featurebyte.atlassian.net/browse/DEV-2139

## Type of Change

## Checklist


hive-udf/lib/src/main/java/com/featurebyte/hive/udf/BaseVectorAggregate.java
```diff
@@ -135,17 +126,28 @@ public void merge(AggregationBuffer agg, Object partial) {
...
-        myagg.container = myList;
+        myagg.container = doubleList;
         return;
       }
 
       // If not, compare the two lists, and update to the max value.
-      assert (container.size() == myList.size());
-      doMerge(container, myList);
+      if (container.size() != doubleList.size()) {
+        throw new RuntimeException(
```
jevonyeoh, line 144: we update this to throw a runtimeexception as assert errors get swallowed unless a parameter is explicitly provided for when running the jar 

https://stackoverflow.com/a/56262724

tests/integration/api/test_vector_aggregation_operations.py
```diff
@@ -57,27 +119,38 @@ def apply_function(xs):
     return out
 
 
-def test_random(data_with_array):
+@pytest.fixture(name="vector_user_entity", scope="module")
```
jevonyeoh, line 122: we recreate a couple of entities here, as otherwise, there's some conflicts with the entity relationship hierarchy that conflicts w/ the non vector event/item tables.

featurebyte/session/spark.py
```diff
@@ -269,26 +270,33 @@ def _process_batch_data(self, data: pd.DataFrame, schema: Schema) -> pd.DataFram
...
             # Convert decimal columns to float
-            if schema.field(i).type == pa.float64() and not is_float_dtype(data[column]):
+            if current_type == pa.float64() and not is_float_dtype(data[column]):
                 data[column] = data[column].astype(float)
-            elif isinstance(schema.field(i).type, pa.TimestampType) and not is_datetime64_dtype(
+            elif isinstance(current_type, pa.TimestampType) and not is_datetime64_dtype(
                 data[column]
             ):
                 data[column] = pd.to_datetime(data[column])
+            elif current_type == pa.list_(pa.float64()):
```
YS-L, line 285: Would it make sense to make `pa.list_(pa.float64())` a constant that is shared between here and above? I assume they have to match exactly.

featurebyte/session/spark.py
```diff
@@ -269,26 +270,33 @@ def _process_batch_data(self, data: pd.DataFrame, schema: Schema) -> pd.DataFram
...
+                # Check if column is string. If so, convert to a list.
+                is_string_series = data[column].apply(lambda x: isinstance(x, str))
+                if is_string_series.any():
```
YS-L, line 288: Curious what is this not the case? If only some of the elements are string, is it fine to apply the eval on all elements but not just the elements of string type?

featurebyte/session/spark.py
```diff
@@ -297,13 +305,15 @@ def _read_batch(self, cursor: Cursor, schema: Schema, batch_size: int = 1000) ->
         pa.RecordBatch
             None if no more rows are available
         """
+        schema = pa.schema(
+            {metadata[0]: self._get_pyarrow_type(metadata[1]) for metadata in cursor.description}
+        )
```
YS-L, line 310: Does the schema change between batches hence the need to reconstruct schema on every batch of data?

tests/integration/api/test_vector_aggregation_operations.py
```diff
@@ -86,37 +159,193 @@ async def register_table_with_array_column(
...
+        skip_fill_na=True,
+    )[feature_name]
+
+    preview_params = {"POINT_IN_TIME": "2022-06-06 00:58:00", "vector_user_id": "2"}
+    feature_preview = feature.preview(pd.DataFrame([preview_params]))
+    assert feature_preview.shape[0] == 1
+    assert feature_preview.iloc[0].to_dict() == {
+        feature_name: expected_results,
+        **convert_preview_param_dict_to_feature_preview_resp(preview_params),
+    }
```
YS-L, line 270: Would be good to also test computing historical features.

featurebyte/session/spark.py
```diff
@@ -2,12 +2,13 @@
 SparkSession class
 """
 # pylint: disable=duplicate-code
+# pylint: disable=wrong-import-order
```
gee-senbong, line 5: Maybe we could just ignore `wrong-import-order` as we use isort.

gee-senbong (approved):
LGTM

featurebyte/session/spark.py
jevonyeoh: hm.. no i don't think it does. i moved this in since it looked duplicated between both callers that were hitting it, but i guess i'll move it back out since it is a bit unnecessary to repeat this work multiple times.

featurebyte/session/spark.py
jevonyeoh: hm it should technically apply only when all of them are strings.

featurebyte/session/spark.py
jevonyeoh: yeap makes sense, will update!

tests/integration/api/test_vector_aggregation_operations.py
jevonyeoh: sounds good, will update too.

tests/integration/api/test_vector_aggregation_operations.py
```diff
@@ -86,37 +160,199 @@ async def register_table_with_array_column(
...
-        "vector_max": "[3,3,3]",
+        feature_name: expected_results,
         **convert_preview_param_dict_to_feature_preview_resp(preview_params),
     }
+
+    feature_list = FeatureList([feature], name="vector_agg_list")
+    observation_set_df = pd.DataFrame([preview_params])
+    historical_features = feature_list.compute_historical_features(observation_set_df)
+    assert historical_features.shape[0] == 1
+    assert list(historical_features.iloc[0]["vector_agg"]) == expected_results
```
jevonyeoh, line 277: @YS-L - i added the test for compute_historical_features here. however, the value is coming back as a numpy array instead of a simple list. i'm going to leave this for now, and fix in a follow-up since there's a workaround if that's ok with you!

YS-L (approved):

PR #1694, opened 2023-08-30
[DEV-2194] implement Context API object

yanhuisg:
## Description

Implement Context API Object

## Type of Change

## Checklist


featurebyte/api/context.py
```diff
@@ -0,0 +1,80 @@
...
+        dict[str, Any]
+        """
+        data = ContextCreate(**self.dict(by_alias=True))
+        return data.json_dict()
+
+    @classmethod
+    def create(
+        cls,
+        name: str,
+        entity_ids: List[PydanticObjectId],
```
jevonyeoh, line 53: do we want to take in entity names, instead of entity IDs? the [spec](https://featurebyte.atlassian.net/wiki/spaces/DEV/pages/16809996/Context#%5CuD83D%5CuDDD2-Requirements) seems to suggest taking in entity names instead, which also seems to be more consistent w/ the rest of the SDK (i.e. users shouldn't have to directly worry about the actual IDs most of the time)

jevonyeoh (commented):
seems like the spec describes a few more endpoints (info, add_eda_observation_set, get_observation_set, update_default_observation_set etc.). guessing we'll add those in at a later point?

featurebyte/api/catalog.py
```diff
@@ -548,6 +549,24 @@ def list_targets(self, include_id: Optional[bool] = True) -> pd.DataFrame:
...
+        include_id: Optional[bool]
+            Whether to include id in the list.
+        Returns
+        -------
+        pd.DataFrame
+            Dataframe of contexts
+        Examples
+        --------
+        >>> contexts = catalog.list_contexts()
+        """
```
gee-senbong, line 567: nit: Would be good to format the doc string here.

featurebyte/api/context.py
```diff
@@ -0,0 +1,80 @@
...
+    _get_schema = ContextModel
+    _update_schema_class = ContextUpdate
+    _list_fields = [
+        "name",
+        "entity_ids",
+        "description",
+    ]
+
+    # pydantic instance variable (public)
+    entity_ids: List[PydanticObjectId]
```
gee-senbong, line 36: If we do the following:
```
context = Context(...).save()

another_context = catalog.get_context_by_id(context.id)
print(another_context.entity_ids)
```
I would expect the output will be empty `entity_ids`. We add add a property `entity_ids` that retrieve it from the `cached_model`.

gee-senbong: Seems like we have the test to cover this. As long as `entity_ids` is immutable after creation, this should be fine.

featurebyte/api/context.py
```diff
@@ -0,0 +1,80 @@
...
+
+        Returns
+        -------
+        dict[str, Any]
+        """
+        data = ContextCreate(**self.dict(by_alias=True))
+        return data.json_dict()
+
+    @classmethod
+    def create(
```
gee-senbong, line 50: Should we add `@typechecked` here? Maybe we should add it for `Entity` api object also.

featurebyte/api/context.py
yanhuisg: good point. I think it should probably using Entity objects instead. will make the change

featurebyte/api/context.py
yanhuisg: ok

gee-senbong (approved):
LGTM

PR #1714, opened 2023-09-04
[DEV-2219] vector functions: add cosine_similarity

jevonyeoh:
## Related Issue
https://featurebyte.atlassian.net/browse/DEV-2219

## Type of Change

## Checklist

### DIFF
--- a/featurebyte/core/accessor/vector.py
+++ b/featurebyte/core/accessor/vector.py
@@ -1,0 +1,70 @@
+"""
+Vector accessor module
+"""
+from __future__ import annotations
+
+from typing import TYPE_CHECKING, TypeVar
+
+from featurebyte.common.doc_util import FBAutoDoc
+from featurebyte.core.util import series_binary_operation
+from featurebyte.enum import DBVarType
+from featurebyte.query_graph.enum import NodeType
+
+if TYPE_CHECKING:
+    from featurebyte.core.series import FrozenSeries
+else:
+    FrozenSeries = TypeVar("FrozenSeries")
+
+
+class VectorAccessorMixin:
+    """
+    VectorAccessorMixin class
+    """
+
+    @property
+    def vec(self: FrozenSeries) -> VectorAccessor:  # type: ignore
+        """
+        Vector accessor object
+
+        Returns
+        -------
+        VectorAccessor
+        """
+        return VectorAccessor(self)
+
+
+class VectorAccessor:
+    """
+    VectorAccessor class used to manipulate vector type FrozenSeries object
+    """
+
+    # documentation metadata
+    __fbautodoc__ = FBAutoDoc(
+        proxy_class="featurebyte.VectorAccessor",
+    )
+
+    def __init__(self, obj: FrozenSeries):
+        if obj.dtype != DBVarType.ARRAY:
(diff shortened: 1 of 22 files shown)


featurebyte/core/accessor/vector.py
```diff
@@ -0,0 +1,70 @@
...
+        Parameters
+        ----------
+        other: FrozenSeries
+            The other vector series
+
+        Returns
+        -------
+        FrozenSeries
+        """
+        return series_binary_operation(
```
YS-L, line 64: Should we validate that the other series is also of vector type? Is it possible to validate that the vectors are of the same dimension?

featurebyte/query_graph/node/vector.py
```diff
@@ -0,0 +1,55 @@
...
+        config: CodeGenerationConfig,
+        context: CodeGenerationContext,
+    ) -> Tuple[List[StatementT], VarNameExpressionInfo]:
+        var_name_expressions = self._assert_no_info_dict(node_inputs)
+        var_name_expression = var_name_expressions[0].as_input()
+        other_operands = [val.as_input() for val in var_name_expressions[1:]]
+        expression = ExpressionStr(
+            f"{var_name_expression}.vec.cosine_similarity(other={other_operands[0]})"
+        )
+        return [], expression
```
YS-L, line 55: Is this code path covered in a test? I think one way to test is to either 1) explicit retrieve and check the generated code or 2) save a feature that involves such an operation and the backend will trigger the code generation and validation automatically end to end.

tests/integration/udf/test_vector_cosine_similarity.py
```diff
@@ -0,0 +1,34 @@
+"""
+Test vector cosine similarity
+"""
+import numpy as np
+import pytest
+
+
+@pytest.mark.parametrize("source_type", ["snowflake"], indirect=True)
```
YS-L, line 8: Should spark be covered in this test as well?

featurebyte/core/accessor/vector.py
jevonyeoh: hm yea i'll add the validation!

do they have to be of the same dimension for cosine similarity? i'm not too familiar w/ how to interpret the values, but the current UDFs will support vectors of different dimensions.

tests/integration/udf/test_vector_cosine_similarity.py
jevonyeoh: oops forgot to add that in here after adding in the spark UDF. will add too

featurebyte/query_graph/node/vector.py
jevonyeoh: @YS-L - i added a basic unit test for this that is similar to some of the other ones in the test file.

featurebyte/core/accessor/vector.py
YS-L: The current implementation implicitly truncates the extra dimensions in the longer vector. I'm not sure what is the desired behaviour here, might want to clarify on that.

YS-L (approved):

featurebyte/core/accessor/vector.py
jevonyeoh: i'll merge this in first and follow-up with changes if necessary!

PR #1801, opened 2023-10-05
[DEV-2399] observation_table: track earliest point in time and entity count

jevonyeoh:
## Description
Track the earliest point in time, and the unique entity counts for entities in the observation table.

## Related Issue
https://featurebyte.atlassian.net/browse/DEV-2399

## Type of Change

## Checklist

### DIFF
--- a/featurebyte/service/observation_table.py
+++ b/featurebyte/service/observation_table.py
@@ -10,7 +10,7 @@ from dataclasses import dataclass
 import pandas as pd
 from bson import ObjectId
 
-from featurebyte import SourceTable
+from featurebyte.api.source_table import SourceTable
 from featurebyte.common.utils import dataframe_from_json
 from featurebyte.enum import DBVarType, MaterializedTableNamePrefix, SpecialColumnName
 from featurebyte.exception import (
@@ -18,8 +18,8 @@ from featurebyte.exception import (
     ObservationTableInvalidContextError,
     UnsupportedPointInTimeColumnTypeError,
 )
-from featurebyte.models import FeatureStoreModel
 from featurebyte.models.base import FeatureByteBaseDocumentModel
+from featurebyte.models.feature_store import FeatureStoreModel
 from featurebyte.models.materialized_table import ColumnSpecWithEntityId
 from featurebyte.models.observation_table import ObservationTableModel
 from featurebyte.persistent import Persistent
--- a/featurebyte/service/observation_table.py
+++ b/featurebyte/service/observation_table.py
@@ -253,7 +253,9 @@ class ObservationTableService(
         column_name_to_count = {}
         for col in entity_cols:
             col_name = col.name
-            column_name_to_count[col_name] = describe_stats_dataframe.loc["unique", col_name]
+            column_name_to_count[col_name] = describe_stats_dataframe.loc[
+                "unique", col_name
+            ]  # pylint: disable=no-member
         return column_name_to_count
 
     async def validate_materialized_table_and_get_metadata(
--- a/featurebyte/service/observation_table.py
+++ b/featurebyte/service/observation_table.py
@@ -25,9 +25,7 @@ from featurebyte.models.observation_table import ObservationTableModel
 from featurebyte.persistent import Persistent
 from featurebyte.query_graph.model.common_table import TabularSource
 from featurebyte.query_graph.node.schema import TableDetails
-from featurebyte.query_graph.sql.materialisation import (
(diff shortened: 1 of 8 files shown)


tests/unit/service/test_observation_table.py
```diff
@@ -128,48 +138,80 @@ async def mock_list_table_schema(*args, **kwargs):
...
-    expected_query = textwrap.dedent(
-        """
-        SELECT
-          MAX("POINT_IN_TIME")
-        FROM "fb_database"."fb_schema"."fb_table"
-        """
-    ).strip()
-    query = db_session.execute_query.call_args[0][0]
-    assert query == expected_query
+        expected_query = textwrap.dedent(
```
jevonyeoh, line 166: the query here has changed to be the last query executed (which is the one to get the describe stats)

featurebyte/service/observation_table.py
```diff
@@ -3,32 +3,136 @@
...
+class PointInTimeStats:
+    """
+    Point in time stats
+    """
+
+    earliest: str
+    most_recent: str
+
+
+async def get_point_in_time_stats(
```
jevonyeoh, line 53: @YS-L - do you think i could replace this call with the `min TZ offset` and `max TZ offset` stats instead?

YS-L: I don't think so since those are min and max for the TZ offset column of the event timestamp, but here we need the min and max for the point in time column.

tests/unit/service/test_observation_table.py
YS-L: Would it make sense to check all the relevant queries ran if there are more than one?

tests/integration/api/test_observation_table.py
```diff
@@ -94,6 +94,8 @@ async def test_observation_table_from_source_table(
     check_materialized_table_preview_methods(
         observation_table, expected_columns=["POINT_IN_TIME", "User ID"]
     )
+    assert observation_table.earliest_point_in_time is not None
+    assert "User ID" in observation_table.entity_column_name_to_count
```
YS-L, line 98: We could make the check stronger by downloading the observation table as a dataframe and check that the stats are expected. Something like:

```python
df = observation_table.to_pandas()
expected_min = df["POINT_IN_TIME"].min()
expected_max = df["POINT_IN_TIME"].max()
for col in entity_cols:
    entity_unique_count = df[col].nunique()
```

featurebyte/service/observation_table.py
```diff
@@ -158,50 +287,30 @@ async def validate_materialized_table_and_get_metadata(
...
+        # Get entity statistics metadata
+        column_name_to_count = await self._get_column_name_to_entity_count(
+            feature_store, table_details, columns_info
+        )
         return {
             "columns_info": columns_info,
             "num_rows": num_rows,
-            "most_recent_point_in_time": most_recent_point_in_time,
+            "most_recent_point_in_time": point_in_time_stats.most_recent,
+            "earliest_point_in_time": point_in_time_stats.earliest,
```
YS-L, line 312: nit: I feel `least_recent_point_in_time` is a more natural counterpart to `most_recent_point_in_time`, but feel free to keep it this way if you don't think so.

YS-L (approved):

featurebyte/service/observation_table.py
jevonyeoh: ah gotcha, what about `top` and `min` stats for the point in time column?

tests/unit/service/test_observation_table.py
jevonyeoh: i couldn't find it in the `call_args` param, let me take a closer look tomorrow.

featurebyte/service/observation_table.py
YS-L: `top` is the most frequent value. I think `max` and `min` stats should work.

tests/unit/service/test_observation_table.py
YS-L: There is a `call_args_list` attribute that records all the calls made.

featurebyte/service/observation_table.py
jevonyeoh: oh indeed, i dunno why i missed the `max` 😓 ok i'll try to swap to those, then all the stats can just be collected in one sql query. (and i won't have to deal with updating the mock too heh)

jevonyeoh: actually i'll fix these in a follow-up so that i can unblock jerome for now!

PR #1908, opened 2023-11-10
[DEV-2527] Update critical data info cleaning operation validation

gee-senbong:
## Description

This PR updates cleaning operation validation by
* trying to cast the imputed value to the assign column type
* checking whether the cleaning operation is supported by the column type

## Type of Change

## Checklist


featurebyte/query_graph/node/cleaning_operation.py
```diff
@@ -225,6 +293,22 @@ class UnexpectedValueImputation(BaseCleaningOperation):
...
+        DBVarType.INT,
+        DBVarType.FLOAT,
+        DBVarType.DATE,
+        DBVarType.TIME,
+        DBVarType.TIMESTAMP,
+        DBVarType.TIMESTAMP_TZ,
+        DBVarType.BOOL,
+        DBVarType.CHAR,
+        DBVarType.VARCHAR,
+    }
```
YS-L, line 306: This set of dtypes seem common and also used in `DisguisedValueImputation`. Would it make sense to define it once and reuse?

tests/unit/query_graph/model/test_column_info.py
```diff
@@ -0,0 +1,185 @@
...
+        (DBVarType.FLOAT, 12, 12.0),
+        (DBVarType.FLOAT, 12.9, 12.9),
+        (DBVarType.FLOAT, "12.9", 12.9),
+        (DBVarType.BOOL, 0, False),
+        (DBVarType.BOOL, "", False),
+        (DBVarType.BOOL, 12.9, True),
+        (DBVarType.DATE, "2023-11-09T11:59:15", "2023-11-09"),
+        (DBVarType.TIMESTAMP, "2023-11-09T11:59:15", "2023-11-09T11:59:15"),
+        (DBVarType.TIMESTAMP, "2023-11-09", "2023-11-09T00:00:00"),
+        (DBVarType.TIMESTAMP, "2021-01-01 00:00:00-0800", "2021-01-01T00:00:00"),
```
YS-L, line 37: It is expected for the timezone information to be dropped in this case? Seems a bit strange to allow specifying a tz-aware CDI value for a tz-naive column.

gee-senbong: We have similar case when casting float number to int (where the number after decimal is truncated). Thinking whether it will be a bit inconsistent if we block this case but ok for another case.

YS-L: For timezone we might want to be more careful since dropping timezone information loses much more information than dropping decimals. If the timestamps in the tz-naive column are in UTC, I think a more reasonable casting process is to convert the tz-aware timestamp to UTC first before dropping the timezone.

gee-senbong: That sounds to be a better solution than failing it. Will update the conversion logic here.

YS-L (approved):

gee-senbong:
Will wait for (https://github.com/featurebyte/featurebyte/pull/1909) to get merged first before merging this PR.

gee-senbong:
Tested this branch locally and it should not introduce additional backward compatibility of existing branch. Will merge this PR first as it is not clear when https://github.com/featurebyte/featurebyte/pull/1909 get merged.

PR #1911, opened 2023-11-12
[DEV-2540] check array and object types for embedding and flat dict

vikua:
## Description

Check and update column attributes for: 
- ARRAY columns being embeddings (potentially)
- OBJECT (snowflake) / STRUCT (spark) columns being flat (non-nested) dicts  

## Related Issue

[DEV-2540](https://featurebyte.atlassian.net/browse/DEV-2540)

## Type of Change

## Checklist

[DEV-2540]: https://featurebyte.atlassian.net/browse/DEV-2540?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/featurebyte/service/column_attributes.py
+++ b/featurebyte/service/column_attributes.py
@@ -9,11 +9,11 @@ from abc import ABC, abstractmethod
 import numpy as np
 import pandas as pd
 
-from featurebyte.api.source_table import SourceTable
 from featurebyte.common.utils import dataframe_from_json
 from featurebyte.enum import ColumnAttribute, DBVarType
 from featurebyte.logging import get_logger
 from featurebyte.models.feature_store import TableModel
+from featurebyte.query_graph.enum import NodeOutputType, NodeType
 from featurebyte.query_graph.model.column_info import ColumnInfo
 from featurebyte.schema.feature_store import FeatureStoreSample
 from featurebyte.service.feature_store import FeatureStoreService
@@ -57,12 +57,25 @@ class ColumnAttributesDetectionService:
         feature_store = await self.feature_store_service.get_document(
             document_id=table.tabular_source.feature_store_id,
         )
-        source_table = SourceTable(
-            feature_store=feature_store,
-            tabular_source=table.tabular_source,
-            columns_info=table.columns_info,
+        graph, node = table.construct_graph_and_node(
+            feature_store_details=feature_store.get_feature_store_details(),
+            table_data_dict=table.dict(by_alias=True),
         )
-        graph, node = source_table.frame.extract_pruned_graph_and_node()
+        columns = [
+            col.name
+            for col in table.columns_info
+            if str(col.dtype)
+            in [DBVarType.ARRAY.value, DBVarType.OBJECT.value, DBVarType.STRUCT.value]
+        ]
+        if not columns:
+            return
+
+        node = graph.add_operation(
+            node_type=NodeType.PROJECT,
+            node_params={"columns": columns},
+            node_output_type=NodeOutputType.FRAME,
+            input_nodes=[node],
+        )
         sample = dataframe_from_json(
             await self.preview_service.sample(
                 FeatureStoreSample(
@@ -111,14 +124,16 @@ class ArrayEmbeddingColumnAttributesDetector(BaseColumnAttributesDetector):
                 series = series[pd.notnull(series)]
                 shapes = series.apply(len)
 
+                # skip if arrays have different shape (number of dimensions)
                 if shapes.unique().ravel().shape != (1,):
                     continue
 
+                # skip if arrays are not 1 dimensional
                 is_1d = series.apply(lambda x: len(np.array(x).shape) == 1)
-
                 if not is_1d.all():
                     continue
 
+                # skip if arrays
                 all_num = series.apply(
                     lambda x: np.all(np.isfinite(pd.to_numeric(x, errors="coerce")))
                 )
--- a/featurebyte/service/column_attributes.py
+++ b/featurebyte/service/column_attributes.py
@@ -1,0 +1,150 @@
+"""Module with extra services for data type info detection like embedding columns or flat (non-nested) dicts"""
+from __future__ import annotations
+
+import json
+from abc import ABC, abstractmethod
+
+import numpy as np
+import pandas as pd
+
+from featurebyte.api.source_table import SourceTable
+from featurebyte.common.utils import dataframe_from_json
+from featurebyte.enum import ColumnAttribute, DBVarType
+from featurebyte.logging import get_logger
+from featurebyte.models.feature_store import TableModel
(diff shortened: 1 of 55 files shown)


featurebyte/service/column_attributes.py
```diff
@@ -0,0 +1,157 @@
...
+        """
+        feature_store = await self.feature_store_service.get_document(
+            document_id=table.tabular_source.feature_store_id,
+        )
+        source_table = SourceTable(
+            feature_store=feature_store,
+            tabular_source=table.tabular_source,
+            columns_info=table.columns_info,
+        )
+        graph, node = source_table.frame.extract_pruned_graph_and_node()
```
gee-senbong, line 65: We could project those affected columns only (for example columns with `ARRAY` or `OBJECT` type).

featurebyte/service/column_attributes.py
```diff
@@ -0,0 +1,157 @@
...
+
+    async def detect(self, columns_info: list[ColumnInfo], sample: pd.DataFrame) -> None:
+        for column in columns_info:
+            if str(column.dtype) == DBVarType.ARRAY.value:
+                series = sample[column.name]
+                series = series[pd.notnull(series)]
+                shapes = series.apply(len)
+
+                if shapes.unique().ravel().shape != (1,):
+                    continue
```
gee-senbong, line 115: Would be good to add a comment here with we continue if the if statement is true.

featurebyte/session/snowflake.py
```diff
@@ -315,6 +316,12 @@ def get_columns_schema_from_dataframe(dataframe: pd.DataFrame) -> list[tuple[str
             ):
                 # Consider the type as an ARRAY if all elements are None, or a list.
                 db_type = "ARRAY"
+            elif (
+                dataframe.shape[0] > 0
+                and dataframe[colname].apply(lambda x: x is None or isinstance(x, dict)).all()
+            ):
+                # Consider the type as an OBJECT if all elements are None, or a dict.
+                db_type = "OBJECT"
```
gee-senbong, line 324: 🤔 Should we remove the condition `dataframe[colname].apply(lambda x: x is None`?

If 
* `dataframe.shape[0] > 0`
* `dataframe[colname].apply(lambda x: x is None` is `True`
line 314-315 will be true.  It `db_type` will be set to `ARRAY`.  It seems to me that the additional check on  `dataframe[colname].apply(lambda x: x is None` is redundant.

tests/integration/conftest.py
```diff
@@ -1093,6 +1107,10 @@ def create_transactions_event_table_from_data_source(
             "ÀMOUNT": "FLOAT",
             "TZ_OFFSET": "VARCHAR",
             "TRANSACTION_ID": "VARCHAR",
+            "EMBEDDING_ARRAY": "ARRAY",
+            "ARRAY": "ARRAY",
+            "FLAT_DICT": "OBJECT" if data_source.type == SourceType.SNOWFLAKE else "STRUCT",
+            "NESTED_DICT": "OBJECT" if data_source.type == SourceType.SNOWFLAKE else "STRUCT",
```
gee-senbong, line 1113: 🤔 Should we introduce `OBJECT` type here or reuse existing `STRUCT` type?

tests/unit/api/test_table.py
```diff
@@ -73,11 +73,14 @@ def test_get_scd_table(saved_scd_table, snowflake_scd_table):
     assert expected_msg in str(exc.value)
 
 
-def test_get_dimension_table(saved_dimension_table, snowflake_dimension_table):
+def test_get_dimension_table(
+    saved_dimension_table, snowflake_dimension_table, mock_add_columns_attributes
+):
```
gee-senbong, line 78: nit: Maybe we could add the `mock_add_columns_attributes` fixture to `snowflake_dimension_table` fixture.

tests/integration/conftest.py
vikua: Snowflake doesn't support STRUCT and treats dicts as OBJECT.  
Spark treats dicts and STRUCT. 
How would we reuse only one? 

We do same for `TIMESTAMP_TZ` don't we? 
https://github.com/featurebyte/featurebyte/blob/vikua/DEV-2540-service/tests/integration/conftest.py#L1099-L1100

featurebyte/service/column_attributes.py
```diff
@@ -0,0 +1,157 @@
...
+
+from typing import Any
+
+import json
+from abc import ABC, abstractmethod
+
+import numpy as np
+import pandas as pd
+
+from featurebyte.api.source_table import SourceTable
```
gee-senbong, line 12: Oh.. we should not import any module from `featurebyte.api` to here.

featurebyte/session/snowflake.py
vikua: Not really. I think this check will still be true if dataframe column contains dicts and None values, right? 
If all values are none - yes, it will be marked as ARRAY, but if some values are none and some are dicts, we still wanna marked it as OBJECT. 
If I remove the check - dict columns with missing values stop being detected as OBJECT. 

Maybe I miss something?

featurebyte/service/column_attributes.py
vikua: Done

tests/integration/conftest.py
gee-senbong: This is the part i am not sure. If `STRUCT` and `OBJECT` have different characteristics (other than due to different feature store type), it makes sense to have two different dtypes. Just want to check whether we should keep one or having two different types. For the `TIMESTAMP` & `TIMESTAMP_TZ`, there are different as one of them have timezone info whereas another one does not have. 

@YS-L Would be good if you can confirm this part.

featurebyte/session/snowflake.py
gee-senbong: You're right. I misread this part as i thought is is something like
```
dataframe[colname].apply(lambda x: x is None)).all() or dataframe[colname].apply(lambda x: isinstance(x, dict)).all()
```

gee-senbong (approved):
LGTM.

tests/integration/conftest.py
vikua: I agree it should probably be something like DICT type for everything to abstract away from DWH details. 
But I'm a bit hesitant to add mapping for STRUCT and OBJECT to new DICT type in this PR. 
Not sure how big the blast radius can be.

YS-L: Got it, I'm fine with the current implementation. Would be good to revisit and clean up if possible.

vikua: https://featurebyte.atlassian.net/browse/DEV-2612 to track

PR #1953, opened 2023-11-29
[DEV-2714] Implement FeatureMaterializeService to prepare features for publishing

YS-L:
## Description

This implements FeatureMaterializeService to prepare features for publishing. It retrieves currently online enabled features that should be materialised together and generates a feature table to be published to offline store. The sql query to generate the feature table is derived from the ingest graphs.

## Type of Change

## Checklist

### DIFF
--- a/featurebyte/service/feature_materialize.py
+++ b/featurebyte/service/feature_materialize.py
@@ -113,6 +113,10 @@ class OnlineEnabledFeaturesMetadata:
     def _get_combined_ingest_graph(self) -> OfflineIngestGraphMetadata:
         """
         Returns a combined ingest graph and related information for all features
+
+        Returns
+        -------
+        OfflineIngestGraphMetadata
         """
         local_query_graph = QueryGraph()
         output_nodes = []
--- a/featurebyte/service/feature_materialize.py
+++ b/featurebyte/service/feature_materialize.py
@@ -1,0 +1,328 @@
+"""
+FeatureMaterializeService class
+"""
+from __future__ import annotations
+
+from typing import List, Optional, Tuple
+
+from dataclasses import dataclass, field
+
+from bson import ObjectId
+from sqlglot import expressions
+from sqlglot.expressions import Select, select
+
+from featurebyte.models.entity import EntityModel
+from featurebyte.models.feature import FeatureModel
+from featurebyte.query_graph.graph import QueryGraph
+from featurebyte.query_graph.model.feature_job_setting import FeatureJobSetting
+from featurebyte.query_graph.node.schema import TableDetails
+from featurebyte.query_graph.sql.adapter import get_sql_adapter
+from featurebyte.query_graph.sql.common import quoted_identifier, sql_to_string
+from featurebyte.query_graph.sql.online_serving import (
+    TemporaryBatchRequestTable,
+    get_online_features,
+)
+from featurebyte.service.entity import EntityService
+from featurebyte.service.feature import FeatureService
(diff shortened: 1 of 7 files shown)


featurebyte/service/feature_materialize.py
```diff
@@ -0,0 +1,332 @@
...
+        if not self.online_store_table_names:
+            return None
+
+        # select distinct serving names across all online store tables
+        distinct_serving_names_from_tables = [
+            select(*[quoted_identifier(serving_name) for serving_name in self.serving_names])
+            .distinct()
+            .from_(table_name)
+            for table_name in self.online_store_table_names
+        ]
```
gee-senbong, line 105: Just to double check my understanding, the online store table here refers to offline store table in feast. What we are going to do here is to get the entity universe of all the offline store tables

tests/integration/feature_store_integration/test_feature_materialize.py
```diff
@@ -0,0 +1,143 @@
...
+    assert set(materialized_features.names) == set(expected[1:])
+    assert df.shape[0] == 5
+
+    # Check offline store table for combined entity (nothing to materialise here)
+    materialized_features = await service.materialize_features(
+        session=session,
+        primary_entity_ids=[user_entity.id, product_action_entity.id],
+        feature_job_setting=default_feature_job_setting,
+    )
+    assert materialized_features is None
```
gee-senbong, line 143: If i understood it correctly, if we add a feature that is grouped by `["ÜSER ID", "PRODUCT_ACTION"]`, we should expect some results here, right?

featurebyte/service/feature_materialize.py
```diff
@@ -0,0 +1,332 @@
...
+        -------
+        OfflineIngestGraphMetadata
+        """
+        local_query_graph = QueryGraph()
+        output_nodes = []
+        output_column_names = []
+
+        for feature in self.features:
+            offline_ingest_graphs = feature.extract_offline_store_ingest_query_graphs()
+            for offline_ingest_graph in offline_ingest_graphs:
```
gee-senbong, line 127: Does it make sense to you if we include `online_store_table_name` in offline ingest query graph? The rationale of doing this is to ease the feast feature view construction.

featurebyte/service/feature_materialize.py
YS-L: The online store tables here refer to our current online store tables, not the ones in feast. Currently the materialisation is based on `get_online_features`, which is batch materialisation using precomputed aggregation results.

featurebyte/service/feature_materialize.py
YS-L: I was thinking of keeping track of that information in a separate collection such as

```python
class OfflineStoreFeatureTable(FeatureByteBaseDocumentModel):

    name: str
    feature_ids: List[PydanticObjectId]
    primary_entity_ids: List[PydanticObjectId]
    # ... and other attributes
```

which will be updated on each feature's enabling / disabling. But I think including the online store table name in offline ingest query graph also makes sense. Which one do you prefer?

In either case, I think we will still have the above collection to simplify `FeatureMaterializeService` and the orchestration of scheduled tasks for feature publishing.

tests/integration/feature_store_integration/test_feature_materialize.py
YS-L: Yes, that is right.

gee-senbong (approved):

PR #1998, opened 2023-12-16
[DEV-2828] Update online serving route to use feast when enabled

YS-L:
## Description

This updates online serving route to use feast to retrieve online features when enabled.

## Type of Change

## Checklist

### DIFF
--- a/Taskfile.yaml
+++ b/Taskfile.yaml
@@ -232,7 +232,9 @@ tasks:
       - featurebyte/**/*
       - tests/**/*
     cmds:
+      - task: test-setup
       - poetry run pytest --reruns=3 --timeout=240 --junitxml=pytest.xml.0 -n auto --cov=featurebyte tests/unit
+      - task: test-teardown
 
   generate-unit-test-fixtures:
     desc: Generate unit test fixtures
--- a/featurebyte/feast/utils/on_demand_view.py
+++ b/featurebyte/feast/utils/on_demand_view.py
@@ -53,10 +53,7 @@ def create_feast_on_demand_feature_view(
         file_handle.write(definition)
 
     module = importlib.import_module(module_name)
-    try:
-        func = getattr(module, function_name)
-    except:
-        raise
+    func = getattr(module, function_name)
     try:
         output = on_demand_feature_view(sources=sources, schema=schema)(func)
     except IOError:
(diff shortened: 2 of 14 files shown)


featurebyte/feast/utils/on_demand_view.py
```diff
@@ -45,7 +46,7 @@ def create_feast_on_demand_feature_view(
     OnDemandFeatureView
         The created OnDemandFeatureView
     """
-    module_name = "on_demand_feature_view"
+    module_name = f"on_demand_feature_view_{ObjectId()}"
```
YS-L, line 49: The same `on_demand_feature_view_dir` is shared across different on demand feature views. If the module name is not unique, subsequent imports of the module has no effect. Making the module name unique is one way to fix the issue.

Taskfile.yaml
```diff
@@ -232,7 +232,9 @@ tasks:
       - featurebyte/**/*
       - tests/**/*
     cmds:
+      - task: test-setup
```
YS-L, line 235: This is to get the local redis up for the online serving tests.

featurebyte/feast/utils/on_demand_view.py
gee-senbong: Ya, i encountered similar issue when writing the test to check ingest graph. I thought temporary directory should fix for this issue, that's why the object id was not introduced in the module name.

tests/fixtures/offline_store_feature_table/feature_cluster_two_features.json
```diff
@@ -140,7 +140,7 @@
                         "table_id": {
                             "$oid": "6337f9651050ee7d5980660d"
                         },
-                        "view_mode": "manual"
+                        "view_mode": "auto"
```
gee-senbong, line 143: Do you know why the view mode is different here? Are we using non-saved features to construct the feature cluster? If the feature is saved, the view mode should be changed to `manual`.

gee-senbong (approved):

tests/fixtures/offline_store_feature_table/feature_cluster_two_features.json
YS-L: Mixing calling API routes and services directly in service unit tests was causing some strange issues such as the deadlock between logging and alive_bar.

To prevent that, I made some changes to the test helper function `deploy_feature` to deploy a feature object using services instead of through the routes. That caused the change in view mode here, though the feature should be saved already:

https://github.com/featurebyte/featurebyte/blob/244466b36914b9a72a7b8bfa2f670ded3303879f/tests/util/helper.py#L625-L653

PR #2005, opened 2023-12-19
[DEV-2797] Add on-demand-feature-view support for count dictionary nodes

gee-senbong:
## Description

This PR implements on-demand-feature-view support for count-dictionary-node-related operations.

## Type of Change

## Checklist

### DIFF
--- a/featurebyte/query_graph/node/count_dict.py
+++ b/featurebyte/query_graph/node/count_dict.py
@@ -2,7 +2,7 @@
 This module contains datetime operation related node classes
 """
 # DO NOT include "from __future__ import annotations" as it will trigger issue for pydantic model nested definition
-from typing import Callable, ClassVar, List, Literal, Optional, Sequence, Set, Tuple, Union
+from typing import Callable, ClassVar, Dict, List, Literal, Optional, Sequence, Set, Tuple, Union
 from typing_extensions import Annotated
 
 import textwrap  # pylint: disable=wrong-import-order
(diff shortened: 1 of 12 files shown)


featurebyte/query_graph/node/count_dict.py
```diff
@@ -119,6 +174,118 @@ def generate_expression(self, operand: str, other_operands: List[str]) -> str:
...
+                cd_var_name, var_name_generator, operation="min"
+            )
+            statements.extend(least_freq_statements)
+            return statements, ExpressionStr(f"{least_freq_var}.reindex({mask_var}.index)")
+        if isinstance(self.parameters, self.UniqueCountParameters):
+            unique_count_statements, unique_count_var = self._get_unique_count(
+                cd_var_name, var_name_generator, self.parameters.include_missing
+            )
+            statements.extend(unique_count_statements)
+            return statements, ExpressionStr(f"{unique_count_var}.reindex({mask_var}.index)")
```
YS-L, line 286: The code here seems a bit dense but there are some repetitions: the only difference is the method that gets called. Would be good if it can be refactored.

featurebyte/query_graph/node/nested.py
```diff
@@ -645,4 +639,11 @@ def _derive_on_demand_view_code(
             var_name = var_name_generator.convert_to_variable_name("feat", node_name=self.name)
             expression = get_object_class_from_function_call("pd.to_datetime", expr)
             return [(var_name, expression)], var_name
+        if self.parameters.output_dtype == DBVarType.OBJECT:
+            var_name = var_name_generator.convert_to_variable_name("feat", node_name=self.name)
+            expression = get_object_class_from_function_call(
+                f"{expr}.apply",
+                ExpressionStr("lambda x: np.nan if pd.isna(x) else ast.literal_eval(x)"),
```
YS-L, line 646: Is `ast.literal_eval` or `json.loads` more appropriate? I'm under the impression that the dictionaries are encoded as json when returned. Curious the behaviour when the data has special key / values like unicode.

tests/integration/feature_store_integration/test_feature_materialize.py
```diff
@@ -77,8 +77,34 @@ def features_fixture(event_table):
...
+        "ÀMOUNT",
+        method="sum",
+        windows=["7d"],
+        feature_names=["amount_sum_across_action_7d"],
+    )[
+        "amount_sum_across_action_7d"
+    ]
+
+    cross_aggregate_feature2 = event_view.groupby(
+        ["CUST_ID"], category="PRODUCT_ACTION"
```
YS-L, line 93: Is there a reason to use a different groupby key here? If using the same key, with different windows the similarity should be non-zero as a sanity check (assuming both windows covered some events).

tests/integration/feature_store_integration/test_feature_materialize.py
```diff
@@ -77,8 +77,34 @@ def features_fixture(event_table):
...
+    )[
+        "amount_sum_across_action_24d"
+    ]
+
+    feature_6 = cross_aggregate_feature1.cd.cosine_similarity(cross_aggregate_feature2)
+    feature_6.name = "EXTERNAL_FS_COSINE_SIMILARITY"
+
     # Save all features to be deployed
-    features = [feature_1, feature_2, feature_3, feature_4, feature_5]
+    features = [feature_1, feature_2, feature_3, feature_4, feature_5, feature_6]
```
YS-L, line 107: Would be good to also check for a count dict feature that is not post-processed into numeric type.

featurebyte/query_graph/node/nested.py
gee-senbong: This is a good catch!

tests/integration/feature_store_integration/test_feature_materialize.py
gee-senbong: The main reason to use a different groupby key here is to make sure it triggers on demand feature view construction. I could revise this test to use the same groupby key but different feature job setting.

gee-senbong: I forget that we block the composite feature with different feature job setting. 😓 
This test actually aims to check the integration part for the count dict related operation. The more detailed tests are covered in the unit tests (`tests/unit/query_graph/odfv/test_count_dict.py`). Do you have any suggestion that we can construct the cosine similarity feature that will trigger ODFV construction?

tests/integration/feature_store_integration/test_feature_materialize.py
gee-senbong: If i understood the request here correctly, do you mean check for a count dict feature that is post-processed into non-numeric type? For post-processed into non-numeric type, I need lookup feature & currently it is not supported (directly calling the method like `key_with_highest_value` or `most_frequent` do not trigger ODFV construction, method like `get_value` or `get_rank` require lookup feature.).

tests/integration/feature_store_integration/test_feature_materialize.py
YS-L: I see. I think it's fine to leave this as is if there are already more detailed unit tests.

tests/integration/feature_store_integration/test_feature_materialize.py
YS-L: Oh actually I meant simply using the dictionaries as features without post-processing.

YS-L (approved):

PR #2018, opened 2023-12-26
[DEV-2871] Add TTL handling for online features

gee-senbong:
## Description

This PR adds TTL handling for online features so that the outdated features are reset after certain period.

## Type of Change

## Checklist


featurebyte/service/feature_materialize.py
```diff
@@ -288,7 +288,7 @@ async def initialize_new_columns(
         materialize_partial(
             feature_store=feature_store,
             feature_view=feature_store.get_feature_view(feature_table_model.name),
-            columns=materialized_features.column_names,
+            columns=materialized_features.column_names + [InternalName.FEATURE_TIMESTAMP_COLUMN],
```
gee-senbong, line 291: @YS-L Currently, the ingest feature timestamp is not in the online feature store. Adding an ingest timestamp column here to populate the timestamp column to feature store. Please let me know if there are better way to handle this.

YS-L: This is done only for `initialize_new_columns()`. Should `scheduled_materialize_features()` also be updated? Should we check whether a feature table require ttl handling before adding this timestamp column?

featurebyte/query_graph/transform/on_demand_view.py
```diff
@@ -97,7 +103,79 @@ def _post_compute(
...
+
+        # variable names
+        req_time_var_name = var_name_generator.convert_to_variable_name(
+            variable_name_prefix="request_time", node_name=None
+        )
+        cutoff_var_name = var_name_generator.convert_to_variable_name(
+            variable_name_prefix="cutoff", node_name=None
+        )
+        ingested_var_name = var_name_generator.convert_to_variable_name(
+            variable_name_prefix="ingested_time", node_name=None
```
YS-L, line 158: I think it would be a bit more consistent to call this `feature_timestamp` since it's referred to that way in the rest of the codebase.

featurebyte/service/feature_materialize.py
gee-senbong: > This is done only for `initialize_new_columns()`. Should `scheduled_materialize_features()` also be updated?
Not sure the reason to update `scheduled_materialize_features`. Would be good if you could clarify this. The main purpose of this update is to populate the online stores with table's ingested timestamp.

```
Should we check whether a feature table require ttl handling before adding this timestamp column?
```
It would be better if we could optimize this part.

YS-L: Is the ingested timestamp never updated after being populated? I thought it should be updated every time the feature is computed (done regularly by `scheduled_materialize_features()`), since it refers to the timestamp at which the feature is computed.

gee-senbong: Ok. I think i got it. Do you mean we should also update `materialize_partial` in `scheduled_materialize_features`?

YS-L: Yes, that's what I meant - otherwise the ingested timestamp in the online store will always be the initial value.

YS-L (approved):

PR #2053, opened 2024-01-05
[DEV-2912] feature table cache metadata service

vikua:
## Description

One of many PRs for feature table cache implementation. 
Metadata storage level - store mediate on Feature Store Tables as well as features cached in them. 

## Related Issue

## Type of Change

## Checklist


featurebyte/service/feature_table_cache.py
```diff
@@ -0,0 +1,153 @@
...
+        observation_table_id: PydanticObjectId
+            Observation table id
+        feature_definitions: List[CachedFeatureDefinition]
+            Feature definitions
+        """
+        features = []
+        for feature_def in feature_definitions:
+            feature_name = feature_def.name
+            if feature_def.definition_hash:
+                feature_name = f"{feature_name}_{feature_def.definition_hash}"
```
YS-L, line 119: Features with different names but the same definition hash are considered identical features and can share the same cached column. Is it intended they all get added to the `features` list? How should the `features` field in `FeatureTableCacheModel` be used?

From caller's perspective, one key information this needs to provide is for each feature definition, is the result already cached, and what is the table name and column name of the cached result.

featurebyte/service/feature_table_cache.py
```diff
@@ -0,0 +1,153 @@
...
+            observation_table = await self.observation_table_service.get_document(
+                document_id=observation_table_id
+            )
+
+            entity_ids = []
+            if observation_table.primary_entity_ids:
+                entity_ids = [str(entity_id) for entity_id in observation_table.primary_entity_ids]
+
+            hasher = hashlib.shake_128()
+            name = f"{observation_table.id}_{','.join(sorted(entity_ids))}".encode("utf-8")
```
YS-L, line 144: Since `observation_table.id` already a unique id, I don't think we need to include `entity_ids` in the hash. In other words there isn't another observation table with the same id but different `entity_ids`.

featurebyte/models/feature_table_cache.py
```diff
@@ -0,0 +1,55 @@
...
+    observation_table_id: PydanticObjectId
+        Observation Table Id
+    name: str
+        Name of the Feature Table Cache. Also used as name of the table in DWH
+    features: List[str]
+        List of feature names which are stored in this feature table cache
+    """
+
+    observation_table_id: PydanticObjectId
+    name: str
```
YS-L, line 31: If this field refers to the table name, I think would be good to rename it to something like `table_name` explicitly.

featurebyte/service/feature_table_cache.py
```diff
@@ -0,0 +1,153 @@
...
+            hasher.update(name)
+            name_hash = hasher.hexdigest(20)
+
+            document = FeatureTableCacheModel(
+                observation_table_id=observation_table_id,
+                name=f"feature_table_cache_{name_hash}",
+                features=features,
+            )
```
gee-senbong, line 152: Wopuld be good if we could store the source feature ID for each feature. It makes our life easier
* if we need to get this information from cache (don't need to parse it from the table name)
* if we need to troubleshoot (check whether the source of the definition hash)

featurebyte/models/feature_table_cache.py
```diff
@@ -0,0 +1,55 @@
...
+    features: List[str]
```
gee-senbong, line 32: Could it be better if we use the schema like
```
features: List[FeatureInfo]
```

where `FeatureInfo` is something like
```
class FeatureInfo:
    name: str
    feature_id: PydanticObjectId
```

vikua: We don't really what it to be based feature id, because the goal is to be able to materialise ideated features before they were actually saved as features

featurebyte/service/feature_table_cache.py
vikua: We can probably only use definition_hash as a feature name. 

I see the flow in a following way: 
- We have `observation_table_id` and feature `definition_hash`. We want to determine if feature is already cached. 
- We lookup  `FeatureTableCacheModel` by `observation_table_id`, and check if `definition_hash` is present in the `features` field. 
- If it does - feature is cached. 
- If it doesn't - we cache the feature in the `table_name` (name field which I will rename), and add feature to `features` list.

vikua: Alternative flow can be to have the model flat, e.g. 

```python 
class FeatureTableCacheModel: 
    table_name: str 
    observation_table_id: ObjectId
    feature_definition_hash: str  
```

I think this approach is a bit safer in terms of 16mb limit, but at the same time I think it will be quite hard to create a FeatureTableCacheModel with so many features so they exceed the limit.

featurebyte/models/feature_table_cache.py
gee-senbong: The main purpose of storing feature ID is to trace the source of the cache. Would be good if we could trace whether it is from saved feature or ideated feature.

vikua: This makes sense. What about the case when multiple features with different names and ids have same definition hash? We were hoping to cache such feature once per observation table, whereas adding feature id might be misleading. 
Do we need to worry about this or is it better to just ignore this case for now?

gee-senbong: It is just for the troubleshooting purpose. We only need to track the feature info when the cache is created, right? Just want to know where the value comes from as feature definition hash has too little information to trace. We could definitely add this later.

vikua: Yes sound good, added the feature id

featurebyte/models/feature_table_cache.py
```diff
@@ -0,0 +1,65 @@
...
+    """
+    Model for feature table cache
+
+    id: PydanticObjectId
+        FeatureTableCache id of the object
+    observation_table_id: PydanticObjectId
+        Observation Table Id
+    table_name: str
+        Name of the Feature Table Cache in DWH (SQL table name).
+    feature_definition_hashes: List[str]
```
YS-L, line 36: Type in the docstring is out of sync.

featurebyte/models/feature_table_cache.py
```diff
@@ -0,0 +1,65 @@
...
+    UniqueValuesConstraint,
+)
+
+
+class CachedFeatureDefinition(FeatureByteBaseModel):
+    """
+    Definition of the feature cached in Feature Table Cache
+    """
+
+    feature_id: PydanticObjectId
```
YS-L, line 22: If the feature is not saved yet e.g. an ideated feature, what should the feature_id be? Should this be optional?

YS-L (approved):

featurebyte/models/feature_table_cache.py
vikua: I wan thinking - ideated feature id (they are also saved into separate collection), but this will probably make it even more misleading. I will make this optional

PR #2060, opened 2024-01-08
[DEV-2912] feature table cache

vikua:
## Type of Change

## Checklist

### DIFF
--- a/featurebyte/service/target_helper/base_feature_or_target_computer.py
+++ b/featurebyte/service/target_helper/base_feature_or_target_computer.py
@@ -3,13 +3,15 @@ Base class for feature or target computer
 """
 from __future__ import annotations
 
-from typing import Any, Callable, Coroutine, Generic, List, Optional, TypeVar, Union
+from typing import Any, Callable, Coroutine, Generic, List, Optional, Set, TypeVar, Union
 
 from abc import abstractmethod
-from dataclasses import dataclass
+from dataclasses import dataclass, replace
 
 import pandas as pd
+from bson import ObjectId
 
+from featurebyte.enum import MaterializedTableNamePrefix
 from featurebyte.models.feature_store import FeatureStoreModel
 from featurebyte.models.observation_table import ObservationTableModel
 from featurebyte.models.parent_serving import ParentServingPreparation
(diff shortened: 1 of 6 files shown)


featurebyte/service/target_helper/base_feature_or_target_computer.py
```diff
@@ -191,3 +179,84 @@ async def compute(
...
+        )
+
+        if len(nodes_hash_dict) < len(validation_parameters.nodes):
+            table_details = output_table_details.copy()
+            table_details.table_name = (
+                f"TEMP_{MaterializedTableNamePrefix.FEATURE_TABLE_CACHE}_{ObjectId()}"
+            )
+        else:
+            table_details = output_table_details.copy()
+            table_details.table_name = feature_cache.table_name
```
YS-L, line 224: We cannot let `self._execute()` create the feature cache table because the table created by `self._execute()` will have feature names as the column names, but the column names in the feature cache table should be based on definition hash. We will need to manage the feature cache table creation separately.

featurebyte/service/target_helper/base_feature_or_target_computer.py
```diff
@@ -191,3 +179,84 @@ async def compute(
...
+            observation_set=observation_set,
+            compute_request=compute_request,
+            output_table_details=table_details,
+            validation_parameters=non_cached_validation_parameters,
+            request_column_names=request_column_names,
+        )
+
+        if len(nodes_hash_dict) < len(validation_parameters.nodes):
+            # run merge into
+            pass
```
YS-L, line 236: After this there will be a step to update the cache metadata right? Maybe it will be cleaner to have a new service that manages physical cache tables in the warehouse that also knows how to interact with FeatureTableCacheService after the table create / merge into is done. Then from here we can call the service to do the work.

featurebyte/service/target_helper/base_feature_or_target_computer.py
```diff
@@ -191,3 +179,84 @@ async def compute(
...
+    async def _compute_observation_table(
+        self,
+        observation_set: ObservationTableModel,
+        compute_request: ComputeRequestT,
+        output_table_details: TableDetails,
+    ) -> None:
+        validation_parameters = await self.get_validation_parameters(compute_request)
+        request_column_names = {col.name for col in observation_set.columns_info}
+
+        feature_cache, nodes_hash_dict = await self.feature_table_cache_service.filter_nodes(
```
YS-L, line 208: The feature table cache workflow is applicable only to newer observation tables with row index column. That can be checked by `observation_set.has_row_index`. Otherwise, it needs to fallback to the non-cached workflow.

featurebyte/service/target_helper/base_feature_or_target_computer.py
vikua: Where does it get the columns names? From nodes? 
Do I need to modify Nodes to change column names? 

This is the part which is very cryptic to me - it is quite hard to understand how some code is translated into the SQL query. 
To me it looks like a set of dataclasses/pydantic models which is then magically executed somehow.

PR #2065, opened 2024-01-09
[DEV-2912] feature table cache

vikua:
## Description

Feature table cache PR#2 
I plan to integrate the service in a separate PR. 

## Related Issue

## Type of Change

## Checklist


featurebyte/service/feature_table_cache.py
```diff
@@ -1,111 +1,367 @@
...
+            feature_store=feature_store,
+            output_table_details=output_table_details,
+            serving_names_mapping=serving_names_mapping,
+            is_feature_list_deployed=is_feature_list_deployed,
+            parent_serving_preparation=parent_serving_preparation,
+        )
+
+        feature_names = [
+            expressions.alias_(
+                node.name,
```
YS-L, line 163: This should use `get_node_output_column_name` to obtain the feature name for each node. `node.name` is the name of the node, not the feature represented by the node.

featurebyte/service/feature_table_cache.py
```diff
@@ -1,111 +1,367 @@
...
+        adapter = get_sql_adapter(db_session.source_type)
+        query = sql_to_string(
+            adapter.create_table_as(
+                table_details=TableDetails(
+                    database_name=db_session.database_name,
+                    schema_name=db_session.schema_name,
+                    table_name=final_table_name,
+                ),
+                select_expr=(
+                    expressions.select(InternalName.TABLE_ROW_INDEX)
```
YS-L, line 178: For consistency, let's quote all the identifiers in the query. So this would be `quoted_identifier(InternalName.TABLE_ROW_INDEX)`.

featurebyte/service/feature_table_cache.py
```diff
@@ -1,111 +1,367 @@
...
+        graph: QueryGraph,
+        cached_nodes: Dict[str, Node],
+        non_cached_nodes: Dict[str, Node],
+        serving_names_mapping: Optional[Dict[str, str]] = None,
+        is_feature_list_deployed: bool = False,
+    ) -> None:
+        # create temporary table with features
+        temp_final_cache_table_name = (
+            f"__TEMP__{MaterializedTableNamePrefix.FEATURE_TABLE_CACHE}_{ObjectId()}"
+        )
```
YS-L, line 208: Is this table dropped later after use?

featurebyte/service/feature_table_cache.py
```diff
@@ -1,111 +1,367 @@
...
+            graph=graph,
+            nodes=non_cached_nodes,
+            serving_names_mapping=serving_names_mapping,
+            is_feature_list_deployed=is_feature_list_deployed,
+        )
 
-        if not documents:
-            return None
+        # alter cached tables adding columns for new features
+        new_columns = [f"FEATURE_{definition_hash}" for definition_hash in non_cached_nodes.keys()]
```
YS-L, line 221: Would be good to have a centralised logic to determine the column name in feature cache table from definition hash. If possible, I think it's even better if this name is determined once and stored at the `FeatureTableCacheMetadataModel` level, so if we later change the naming logic we won't break backward compatibility.

featurebyte/service/feature_table_cache.py
```diff
@@ -1,111 +1,367 @@
...
+        alter_expr = expressions.AlterTable(
+            this=expressions.Table(this=quoted_identifier(cache_metadata.table_name)),
+            actions=[
+                expressions.ColumnDef(
+                    this=quoted_identifier(column),
+                    kind=expressions.DataType.build("?"),
```
YS-L, line 227: This function can be used to determine the dtype of the feature: https://github.com/featurebyte/featurebyte/blob/6ddfecc1ab9c0ba8f01737eea661e46ac84d7500/featurebyte/query_graph/transform/offline_store_ingest.py#L73-L75

Then we need to use adapter's `get_physical_type_from_dtype` method to convert it to the DataType here.

featurebyte/service/feature_table_cache.py
```diff
@@ -1,111 +1,367 @@
...
+        )
+        sql = sql_to_string(alter_expr, source_type=db_session.source_type)
+        await db_session.execute_query(sql)
 
-        return documents[0]
+        # merge temp table into cache table
+        merge_conditions = [
+            expressions.EQ(
+                this=get_qualified_column_identifier(
+                    InternalName.TABLE_ROW_INDEX, "faeture_table_cache"
```
YS-L, line 239: Typo in `faeture_table_cache`. Would be good if this is made a constant variable.

featurebyte/service/feature_table_cache.py
```diff
@@ -1,111 +1,367 @@
...
+                expressions.When(
+                    this=expressions.Column(this=expressions.Identifier(this="MATCHED")),
+                    then=update_expr,
+                ),
+                expressions.When(
+                    this=expressions.Not(
+                        this=expressions.Column(this=expressions.Identifier(this="MATCHED")),
+                    ),
+                    then=insert_expr,
+                ),
```
YS-L, line 287: Since the row index columns between the two tables always align, we don't need to have the `NOT MATCHED` branch.

featurebyte/service/feature_table_cache.py
```diff
@@ -1,111 +1,367 @@
...
+        sql = sql_to_string(merge_expr, source_type=db_session.source_type)
+        await db_session.execute_query(sql)
 
-    async def update_feature_table_cache(
+    async def create_or_update_feature_table_cache(
         self,
-        observation_table_id: PydanticObjectId,
-        feature_definitions: List[CachedFeatureDefinition],
+        feature_store: FeatureStoreModel,
+        observation_table: ObservationTableModel,
```
YS-L, line 296: Would be good to have an assertion that `observation_table.has_row_index` must be True when calling this function.

featurebyte/service/feature_table_cache.py
```diff
@@ -1,111 +1,367 @@
...
+        cached_nodes, non_cached_nodes = await self.split_cached_and_non_cached_nodes(
+            cache_metadata, graph, nodes
+        )
+
+        db_session = await self.session_manager_service.get_feature_store_session(
+            feature_store=feature_store
+        )
+        if feature_table_cache_exists:
+            # if feature table cache exists - update existing table with new features
+            if non_cached_nodes:
```
YS-L, line 344: We can return early without creating the session if feature table cache exists and there is non-cached nodes.

featurebyte/service/feature_table_cache.py
```diff
@@ -1,111 +1,367 @@
...
-        Optional[FeatureTableCacheModel]
-            Feature Table Cache model if exists
+        Tuple[Dict[str, str], Dict[str, str]]
+            Tuple of
+            - cached definition hases to nodes mapping
+            - nont cached definition hases to nodes mapping
         """
-        documents = []
+        node_names = [node.name for node in nodes]
+        pruned_graph, node_name_map = graph.quick_prune(target_node_names=node_names)
```
YS-L, line 86: @gee-senbong Is this the right way of pruning the graph before extracting feature definition hashes, combined with the `prepare_graph_to_store` steps below?

gee-senbong: Opps, I missed this comment. The current flow seems correct to me.

gee-senbong (approved):
LGTM

featurebyte/service/feature_table_cache.py
gee-senbong: Just to clarify the logic here:

1. Quick prune here is to remove all used nodes from the input graph.
2. The following statement is used to prune the graph & clean up the view graph node metadata (same flow used to generate graph for feature definition, https://github.com/featurebyte/featurebyte/blob/63e1f00fddf4a426c059e24ec5009d0ff95aa538/featurebyte/worker/util/batch_feature_creator.py#L227-L230).
```
            (
                prepared_graph,
                prepared_node_name,
            ) = await self.namespace_handler.prepare_graph_to_store(
                graph=pruned_graph,
                node=pruned_graph.get_node_by_name(node_name_map[node.name]),
                sanitize_for_definition=True,
            )
```

Since the graph used for feature definition should the consistent with the graph that is saved in persistent (otherwise, feature definition should fail and user can't save the feature), it should be valid that is the final version of the feature's graph.

featurebyte/service/feature_table_cache.py
```diff
@@ -1,111 +1,448 @@
...
-        )
-        self.observation_table_service = observation_table_service
+        self.feature_table_cache_metadata_service = feature_table_cache_metadata_service
+        self.namespace_handler = namespace_handler
+        self.session_manager_service = session_manager_service
+        self.entity_validation_service = entity_validation_service
+        self.tile_cache_service = tile_cache_service
+        self.feature_list_service = feature_list_service
+
+    def create_feature_name(self, definition_hash: str) -> str:
```
YS-L, line 102: Is this referring to the column name in the cache table, not really the name of any features? I feel it is less confusing if we call this `column_name` instead of `feature_name`.

gee-senbong: Besides, would prefer to move this logic to the model level rather than service level as:
* it is easier to test (without the need to construct the service)
* personally, i would search the models first to check where the name of the object is constructed rather than in service level modules.

featurebyte/service/feature_table_cache.py
```diff
@@ -1,111 +1,448 @@
...
+            ]
+            update_expr = expressions.Update(
+                expressions=[
+                    expressions.EQ(
+                        this=get_qualified_column_identifier(
+                            definition.feature_name, merge_target_table_alias
+                        ),
+                        expression=get_qualified_column_identifier(
+                            definition.feature_name, merge_source_table_alias
+                        ),
```
YS-L, line 330: Currently `_update_table()` will create two tables: one created by `get_historical_features`, one created by renaming the original feature names to cache table column names. Both are dropped in the end.

I think it can be improved by only creating only one table: don't create the second table for renaming purpose, and instead use the first table for the merge operation. Update `update_expr` here to account for that (specify the original feature name before renaming).

Not sure the impact, but maybe something we can revisit later.

tests/integration/service/test_feature_table_cache.py
```diff
@@ -0,0 +1,293 @@
...
+from sqlglot import parse_one
+
+from featurebyte import FeatureList
+from featurebyte.enum import InternalName
+from featurebyte.query_graph.sql.common import sql_to_string
+from tests.util.helper import create_observation_table_from_dataframe
+
+
+@pytest.fixture(name="event_view")
+def event_view_fixture(event_table):
```
YS-L, line 17: Are these fixtures copied from the event view integration tests? Would be good to move them to conftest to avoid duplicating code unnecessarily.

featurebyte/service/feature_table_cache.py
vikua: moving to model is a good idea 👍

YS-L (approved):

PR #2070, opened 2024-01-13
[DEV-2912] feature table cache service usage

vikua:
## Description

- Creating view from cache table 
- Using feature table cache in historical features 
- Using feature table cache in target computation 

## Related Issue

## Type of Change

## Checklist

### DIFF
--- a/featurebyte/service/feature_table_cache.py
+++ b/featurebyte/service/feature_table_cache.py
@@ -1,7 +1,7 @@
 """
 Module for managing physical feature table cache as well as metadata storage.
 """
-from typing import Dict, List, Optional, Tuple, cast
+from typing import Any, Callable, Coroutine, Dict, List, Optional, Tuple, cast
 
 from bson import ObjectId
 from sqlglot import expressions
@@ -21,6 +21,7 @@ from featurebyte.query_graph.node.schema import TableDetails
 from featurebyte.query_graph.sql.adapter import get_sql_adapter
 from featurebyte.query_graph.sql.common import (
     get_dialect_from_source_type,
+    get_fully_qualified_table_name,
     get_qualified_column_identifier,
     quoted_identifier,
     sql_to_string,
@@ -30,7 +31,7 @@ from featurebyte.query_graph.transform.offline_store_ingest import extract_dtype
 from featurebyte.service.entity_validation import EntityValidationService
 from featurebyte.service.feature_list import FeatureListService
 from featurebyte.service.feature_table_cache_metadata import FeatureTableCacheMetadataService
-from featurebyte.service.historical_features import get_historical_features
+from featurebyte.service.historical_features_and_target import get_historical_features, get_target
 from featurebyte.service.namespace_handler import NamespaceHandler
 from featurebyte.service.session_manager import SessionManagerService
 from featurebyte.service.tile_cache import TileCacheService
@@ -99,6 +100,46 @@ class FeatureTableCacheService:
         self.tile_cache_service = tile_cache_service
         self.feature_list_service = feature_list_service
 
+    async def nodes_to_defition_hashes(
+        self,
+        graph: QueryGraph,
+        nodes: List[Node],
+    ) -> Dict[str, Node]:
+        """
+        Compute definition caches for list of nodes
+
+        Parameters
+        ----------
+        graph: QueryGraph
+            Graph definition
+        nodes: List[Node]
+            Input node names
+
+        Returns
+        -------
+        Dict[str, Node]
+            Mapping from definition hash to a node
+        """
+        node_names = [node.name for node in nodes]
+        pruned_graph, node_name_map = graph.quick_prune(target_node_names=node_names)
+
+        hashes = {}
+        for node in nodes:
+            (
+                prepared_graph,
+                prepared_node_name,
+            ) = await self.namespace_handler.prepare_graph_to_store(
+                graph=pruned_graph,
+                node=pruned_graph.get_node_by_name(node_name_map[node.name]),
+                sanitize_for_definition=True,
+            )
+            definition_hash_extractor = DefinitionHashExtractor(graph=prepared_graph)
+            definition_hash = definition_hash_extractor.extract(
+                prepared_graph.get_node_by_name(prepared_node_name)
+            ).definition_hash
+            hashes[definition_hash] = node
+        return hashes
+
     async def get_non_cached_nodes(
         self,
         feature_table_cache_metadata: FeatureTableCacheMetadataModel,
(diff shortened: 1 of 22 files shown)


featurebyte/service/feature_table_cache.py
```diff
@@ -99,32 +100,25 @@ def __init__(
         self.tile_cache_service = tile_cache_service
         self.feature_list_service = feature_list_service
 
-    async def get_non_cached_nodes(
+    async def nodes_to_defition_hashes(
```
YS-L, line 103: Should the method name be `definition_hashes_to_nodes` if the return type is `Dict[str, Node]`?

featurebyte/service/target_helper/compute_target.py
```diff
@@ -51,49 +57,30 @@ async def execute(  # pylint: disable=too-many-locals
...
+            and executor_params.observation_set.has_row_index
+        ):
+            await self.feature_table_cache_service.create_view_from_cache(
+                feature_store=executor_params.feature_store,
+                observation_table=executor_params.observation_set,
+                graph=executor_params.graph,
+                nodes=executor_params.nodes,
+                output_view_details=executor_params.output_table_details,
+                serving_names_mapping=executor_params.serving_names_mapping,
+            )
```
YS-L, line 71: Should this set `is_target=True`? Might want to make that parameter non-optimal if it's easy to miss.

featurebyte/service/feature_table_cache.py
```diff
@@ -443,11 +504,105 @@ async def create_or_update_feature_table_cache(
...
+
+        columns_expr = [
+            expressions.alias_(
+                quoted_identifier(cast(str, cached_features[definition_hash])),
+                alias=graph.get_node_output_column_name(node.name),
+                quoted=True,
+            )
+            for definition_hash, node in hashes.items()
+        ]
+        select_expr = expressions.select(*columns_expr).from_(
```
YS-L, line 590: We should include the row index column `InternalName.TABLE_ROW_INDEX` in the view.

YS-L (commented):
Would be good to have some end to end integration test cases (using SDK API) to cover the new flow. E.g. materializing the same observation table on the same features twice. We should be able to check that the materialized features are the same, and ideally the second materialization completes faster.

vikua:
@YS-L do we have any existing integration tests for this functionality?

YS-L:
Yes, for example: https://github.com/featurebyte/featurebyte/blob/d2977c6463cb70d320495a36096d9fc57d1d1ca2/tests/integration/api/test_event_view_operations.py#L476

But there isn't one that will specifically trigger the caching logic.

featurebyte/service/feature_table_cache.py
```diff
@@ -443,11 +507,112 @@ async def create_or_update_feature_table_cache(
...
+                )
+            )
+        )
+        create_view_exr = expressions.Create(
+            this=expressions.Table(
+                this=get_fully_qualified_table_name(output_view_details.dict()),
+            ),
+            kind="VIEW",
+            expression=select_expr,
+            replace=True,
```
YS-L, line 615: Is there a need to set replace? If not, probably safer to not set it to avoid overwriting unexpectedly.

tests/integration/api/test_event_view_operations.py
```diff
@@ -729,6 +741,122 @@ def assert_datetime_almost_equal(s1: pd.Series, s2: pd.Series):
...
+    )
+
+    historical_feature_table_name = f"historical_feature_table_{ObjectId()}"
+    historical_feature_table = feature_list_1.compute_historical_feature_table(
+        observation_table,
+        historical_feature_table_name,
+    )
+    df_historical_features_1 = await get_dataframe_from_materialized_table(
+        session, historical_feature_table
+    )
```
YS-L, line 801: Can we also call `historical_feature_table.to_pandas()` to make sure downloading the table works? Note it's expected that the result in `to_pandas()` doesn't include the row index column.

tests/integration/api/test_event_view_operations.py
```diff
@@ -729,6 +741,122 @@ def assert_datetime_almost_equal(s1: pd.Series, s2: pd.Series):
     assert (s1 - s2).dt.total_seconds().abs().max() <= 1e-6
 
 
+@pytest.mark.parametrize("input_format", ["table", "uploaded_table"])
```
YS-L, line 744: I think covering the "table" input format here is probably sufficient.

YS-L (approved):
LGTM when tests are green. Great work!

featurebyte/session/base.py
```diff
@@ -833,6 +855,19 @@ def remove_materialized_tables(cls, table_names: list[str]) -> list[str]:
...
+        List views in the working schema that can be dropped without losing data. These are the
+        views that will be reinstated by WorkingSchemaService when recreating the working schema.
+
+        Returns
+        -------
+        list[str]
+        """
+        views = await self.session.list_views(self.session.database_name, self.session.schema_name)
+        views_names = [view.name for view in views]
+        return self.remove_materialized_tables(views_names)
```
YS-L, line 869: Why do we need this? So far all the views we create are for historical feature tables, and we don't want to be dropping such views.

featurebyte/worker/task/target_table.py
```diff
@@ -116,7 +114,8 @@ async def execute(self, payload: TargetTableTaskPayload) -> Any:
                 request_input=payload.request_input,
                 primary_entity_ids=primary_entity_ids,
                 purpose=purpose,
-                has_row_index=True,
```
YS-L, line 119: On second thought, I think we want this to be True since we want the created observation table to be able to benefit from caching when used. We should still call the `add_row_index_column` function above when `observation_set` is not a row index eligible observation table.

featurebyte/session/base.py
vikua: Following is happening: 
- Some integration tests are triggering https://github.com/featurebyte/featurebyte/blob/2c0433366d9308a9fbf92dea11d47906e6dac92d/featurebyte/service/working_schema.py#L32 
- This lead (based on the DWH) lead to dropping tables, e.g. for Snowflake this code https://github.com/featurebyte/featurebyte/blob/2c0433366d9308a9fbf92dea11d47906e6dac92d/featurebyte/session/snowflake.py#L405 
- Both snowflake and spark return views as a part of `list_tables` response, but neither can drop view via "DROP TABLE" statement. 
- This lead to failures like "DROP TABLE expects object Table but View received" or something like that.  

That's why I add new function, which drops views separately. 

Regarding whether we want to drop historical feature tables or not - this is a good question. Historical tables are definitively getting dropped right now (being tables). Implementation in master just lists all tabes in schema as far as I can see, it is not very picky on the semantic meaning of the table.

featurebyte/worker/task/target_table.py
vikua: I will add it

featurebyte/session/base.py
vikua: We seem to drop everything except tables which start with `MaterializedTableNamePrefix` prefix. 
Approach to creating a historical feature table name hasn't changed 
https://github.com/featurebyte/featurebyte/blob/2c0433366d9308a9fbf92dea11d47906e6dac92d/featurebyte/worker/task/historical_feature_table.py#L58

but in some reasons it wanted to drop views

vikua: I guess I need to investigate this a bit more. Looking at the code - those views aren't supposed to be dropped.

vikua: Solved, turned out this test have seen some other views created by feature table cache test and was trying to drop those.

PR #2085, opened 2024-01-22
[DEV-2964] target table cache conflicting columns

vikua:
## Description

Observation table with target added has identical column name to the target node in the graph. 
This leads to `get_target` function failing generating SQL with two columns having same name. 

Assuming that they are identical, not adding target column to the cache as it will be added via caching.  

## Related Issue

https://featurebyte.atlassian.net/browse/DEV-2964

## Type of Change

## Checklist

### DIFF
--- a/featurebyte/service/historical_features_and_target.py
+++ b/featurebyte/service/historical_features_and_target.py
@@ -298,6 +298,10 @@ async def get_target(
     request_table_name = f"{REQUEST_TABLE_NAME}_{request_id}"
     request_table_columns = observation_set.columns
 
+    # filter request columns if nodes request them as well
+    nodes_columns = [graph.get_node_output_column_name(node.name) for node in nodes]
+    request_table_columns = [col for col in request_table_columns if col not in nodes_columns]
+
     # Execute feature SQL code
     await observation_set.register_as_request_table(
         session,
--- a/tests/integration/service/test_feature_table_cache.py
+++ b/tests/integration/service/test_feature_table_cache.py
@@ -10,6 +10,8 @@ from sqlglot import parse_one
 
 from featurebyte import FeatureList
 from featurebyte.enum import InternalName
+from featurebyte.models.materialized_table import ColumnSpecWithEntityId
+from featurebyte.models.observation_table import ObservationTableModel
 from featurebyte.query_graph.node.schema import TableDetails
 from featurebyte.query_graph.sql.common import sql_to_string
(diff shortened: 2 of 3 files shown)


tests/integration/service/test_feature_table_cache.py
```diff
@@ -237,6 +239,118 @@ async def test_update_feature_table_cache(
...
+    source_type,
+    persistent,
+):
+    """Test target table cache"""
+    feature_store_model = await feature_store_service.get_document(document_id=feature_store.id)
+    observation_table_model = await observation_table_service.get_document(
+        document_id=observation_table.id
+    )
+
+    # add target column in a hacky way - there is no good way to add target from the integration test
```
YS-L, line 262: Why is that? Is it possible to write the test using only SDK to demonstrate an actual workflow where users will encounter this issue?

vikua: No it is not possible. I didn't find the way to add target values to observation table other than via UI/API -> worker job. Workers are not available in integration tests as I know, please correct me if I'm wrong

featurebyte/service/historical_features_and_target.py
```diff
@@ -298,6 +298,10 @@ async def get_target(
     request_table_name = f"{REQUEST_TABLE_NAME}_{request_id}"
     request_table_columns = observation_set.columns
 
+    # filter request columns if nodes request them as well
+    nodes_columns = [graph.get_node_output_column_name(node.name) for node in nodes]
+    request_table_columns = [col for col in request_table_columns if col not in nodes_columns]
```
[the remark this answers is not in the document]
vikua: I think this issue can occur for any observation table which happen to have a column which have same name as a feature. 

For features it is less common and only possible if users will explicitly add some column which will repeat the name of the feature, and can probably be classified as user error. 

For target it is a bit more tricky. We have a functionality to "Compute New Observation table with target value", i.e. create observation table by adding target column to an existing observation table. 
In this case it cannot be classified as user error as user does not explicitly add any columns to observation table, this happens behind the scenes. 

Regarding caching, Feature Table Cache is designed around definition hash of nodes now, all the caching is based around graph and nodes. 
Columns in observation table are not available as nodes, we operate on them separately as a list of `request_table_columns` (at least to my knowledge). 
Thus, if I decide to skip a target `node`, I have **nothing to cache**, because for target we have only one node - target node - which we what to skip.

PR #2088, opened 2024-01-23
[DEV-2960] Update offline store table naming logic

gee-senbong:
## Description

This PR updates offline store table naming logic so that it generates table name with less than 64 characters.

## Type of Change

## Checklist


featurebyte/query_graph/transform/offline_store_ingest.py
```diff
@@ -251,13 +197,13 @@ def _prepare_offline_store_ingest_query_specific_node_parameters(
...
-        entity_id_to_serving_name: Dict[PydanticObjectId, str],
-        catalog_id: PydanticObjectId,
+        operation_structure: OperationStructure,
     ) -> Dict[str, Any]:
         agg_nodes_info = []
         feature_job_settings = []
+        agg_node_names = [agg.node_name for agg in operation_structure.iterate_aggregations()]
         for node_name in aggregation_node_names:
-            if node_name in node_name_to_subgraph_node_name:
+            if node_name in agg_node_names:
```
gee-senbong, line 206: Update the condition here so that we use only those aggregation node specified in operation structure.

featurebyte/query_graph/transform/decompose_point.py
```diff
@@ -182,16 +182,10 @@ def extract_from_target_node(self, node: Node) -> Optional[FeatureJobSetting]:
...
             if isinstance(_node, AggregationOpStructMixin):
-                fjs = self.extract_from_agg_node(node=_node)
-                if fjs:
-                    feature_job_settings.add(fjs)
-
-        if len(feature_job_settings) > 1:
-            assert False, "Multiple feature job settings found"
-        return next(iter(feature_job_settings), None)
+                return self.extract_from_agg_node(node=_node)
+        return None
```
gee-senbong, line 188: Update the logic here so that only the first feature job setting from the aggregation node is considered.

featurebyte/service/feature_offline_store_info.py
```diff
@@ -0,0 +1,238 @@
...
+            primary_entity_ids=primary_entity_ids,
+            feature_job_setting=feature_job_setting,
+            has_ttl=has_ttl,
+            feature_store_id=catalog.default_feature_store_ids[0],
+            catalog_id=catalog_id,
+            entity_id_to_serving_name=entity_id_to_serving_name,
+        )
+        persist_table = await self.offline_store_feature_table_service.get_or_create_document(
+            data=table
+        )
```
YS-L, line 91: Does this mean offline store feature tables will be created here and no longer by `OfflineStoreFeatureTableManagerService`? I think this change has some impact, for example table comments are added by `OfflineStoreFeatureTableManagerService` when it creates the tables, but that is not done here.

Is it possible to determine the table name here but defer the creation of the tables to `OfflineStoreFeatureTableManagerService`?

gee-senbong: It will create a table without any `feature_ids`. To determine the table name, we need to add records to the persistent. If the table comment is the only concern, could we update the logic to add table comments? It seems to be easier to fix than deferring table creation.

gee-senbong: The only way that i can think of to defer creation is to use the table signature. Based on our previous discussion, it seems to have big impact on current code changes...

gee-senbong: Could i tag a new table here https://github.com/featurebyte/featurebyte/blob/d9438b829743fb83705484e77d016ab3c8be044e/featurebyte/service/offline_store_feature_table_manager.py#L201 if the `feature_ids` of the original table is empty?

YS-L: Updating the table comments logic makes sense to me. That is the only issue I can see now.

Going this route, I think we should also update `OfflineStoreFeatureTableManagerService` to remove the code here:

https://github.com/featurebyte/featurebyte/blob/401c941b14af0cdd3e7a8dcac46ee1707a69264e/featurebyte/service/offline_store_feature_table_manager.py#L208-L220

since it will never be executed.

gee-senbong: Do we need to handle the case where the feature created before this change? If yes, maybe we should keep the code here...

YS-L: With the current implementation I don't feel comfortable for `offline_store_feature_table_service.create_document` to be called anywhere except by `OfflineStoreInfoInitializationService`, since the created table's name can be completely different from the provided `feature_table_model`. I think that can lead to inconsistency and probably not worth supporting.

gee-senbong: As long as it is ok to drop the support for the older feature, i am ok to remove that part.

featurebyte/models/offline_store_feature_table.py
```diff
@@ -43,8 +48,119 @@ class OfflineStoreFeatureTableModel(FeatureByteCatalogBaseDocumentModel):
...
+        ValueError
+            If entity_universe is not set
+        """
+        if self.internal_entity_universe is None:
+            raise ValueError("entity_universe is not set")
+        return EntityUniverseModel(**self.internal_entity_universe)
+
+    def _get_basename(self) -> str:
+        # max length of feature table name is 64
+        # reserving 7 characters for prefix (catalog name, `cat<num>_`, num is 1-999)
```
YS-L, line 117: Looking at the generated table names, I wonder if it would be clearer to spell out `cat` as `catalog`.

YS-L (approved):

PR #2093, opened 2024-01-24
[DEV-2985] Add special handling for features with composite entity

YS-L:
## Description

This adds special handling for features with composite entity so that they work with feast which doesn't support entity with multiple join keys. In the offline store feature tables, serving names of the composite entity are concatenated as a new column which will be used as the entity's join key.

## Type of Change

## Checklist

### DIFF
--- a/featurebyte/service/online_serving.py
+++ b/featurebyte/service/online_serving.py
@@ -304,6 +304,21 @@ class OnlineServingService:  # pylint: disable=too-many-instance-attributes
         - Add point in time column if required
         - Composite entities are combined into a single column
         - Remove columns that are not required since feast is strict and will complain about them
+
+        Parameters
+        ----------
+        request_data : List[Dict[str, Any]]
+            Request data to be processed
+        feature_ids : List[PydanticObjectId]
+            List of feature ids in the feature list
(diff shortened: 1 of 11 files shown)


featurebyte/service/online_serving.py
```diff
@@ -190,21 +208,18 @@ async def get_online_features_by_feast(
...
-            added_column_names = None
+            added_column_names = []
 
-        # Include point in time column if it is required
-        is_point_in_time_column_required = self._require_point_in_time_request_column(
-            feature_cluster
+        request_data_metadata = await self._process_request_columns(
+            request_data=request_data,
+            feature_ids=feature_list.feature_ids,
+            feature_cluster=feature_list.feature_clusters[0],
```
gee-senbong, line 216: Would it be better to retrieve the `feature_cluster` by finding those cluster which matches the feature store ID?

gee-senbong (approved):

featurebyte/service/online_serving.py
YS-L: Hmm I think that might be a bit of an overkill for now since we have the assumption that there is only a single feature cluster in a feature list across the codebase.

PR #2116, opened 2024-02-02
[DEV-2998] Implement async workflow for catalog online store update

YS-L:
## Description

This implements the async workflow for catalog online store update. This is to ensure that the online store is populated with features before being used for online features retrieval.

## Type of Change

## Checklist


featurebyte/routes/catalog/api.py
```diff
@@ -57,7 +58,7 @@ def __init__(self) -> None:
             status_code=HTTPStatus.OK,
         )
 
-        # update online store route
+        # update online store route (for backward compatibility; to be deprecated)
```
kchua78, line 61: I don't think we need to keep this? The update online store function is not in the latest released SDK

featurebyte/service/catalog.py
```diff
@@ -85,6 +88,32 @@ async def update_online_store(
...
+        document_id: ObjectId
+            Catalog id
+        data: CatalogOnlineStoreUpdate
+            Online store update payload
+
+        Returns
+        -------
+        CatalogOnlineStoreInitializeTaskPayload
+        """
+        document = await self.get_document(document_id=document_id)
```
kchua78, line 110: Should we check that that `document.online_store_id` is different from `data.online_store_id` here to avoid launching an unnecessary task?

featurebyte/routes/catalog/api.py
YS-L: This is mainly not to break the app frontend. Or is this route not use there yet?

featurebyte/service/catalog.py
YS-L: Sure, will add this check.

featurebyte/routes/catalog/api.py
kchua78: I see that makes sense 👍

kchua78 (approved):
LGTM

featurebyte/routes/catalog/api.py
```diff
@@ -66,6 +67,14 @@ def __init__(self) -> None:
             status_code=HTTPStatus.OK,
         )
 
+        # update online store route
+        self.router.add_api_route(
+            "/{catalog_id}/online_store_async",
```
gee-senbong, line 72: 🤔 Our current async function does not have `async` in the endpoint.

gee-senbong: Understand this is for backward compatibility.. But it introduces some inconsistencies of existing routes naming convention.

YS-L: Sure, I'm open to suggestions.

YS-L: I plan to rename `online_store_async` back to `online_store` once UI transitioned, so this should be temporary.

gee-senbong: I think frontend is using status code to check whether it is an async task or not. Thinking whether it is ok to just replace existing sync patch endpoint.

featurebyte/worker/task/catalog_online_store_update.py
```diff
@@ -0,0 +1,98 @@
...
+        self.feature_materialize_service = feature_materialize_service
+        self.offline_store_feature_table_service = offline_store_feature_table_service
+        self.feast_registry_service = feast_registry_service
+        self.feast_feature_store_service = feast_feature_store_service
+        self.catalog_service = catalog_service
+
+    async def get_task_description(self, payload: CatalogOnlineStoreInitializeTaskPayload) -> str:
+        if payload.online_store_id is not None:
+            return f'Initialize online store "{payload.online_store_id}" for catalog {payload.catalog_id}'
+        return f"Disabling online store for catalog {payload.online_store_id}"
```
gee-senbong, line 54: Is this typo? Should it be something like
```Disabling online store {...online_store_id} for catalog {...catalog_id}```?

File: featurebyte/worker/task/catalog_online_store_update.py [hunk 5]
```diff
@@ -0,0 +1,98 @@
...
+            )
+            if session is None:
+                session = await self.feature_materialize_service._get_session(  # pylint: disable=protected-access
+                    feature_table_model
+                )
+            await self.feature_materialize_service.update_online_store(
+                feature_store=feast_feature_store,
+                feature_table_model=feature_table_model,
+                session=session,
+            )
```
gee-senbong, line 98: If the task is expect to run for some time, would be good to call `self.task_progress_updater.update_progress` to update progress.

gee-senbong (approved):

featurebyte/routes/catalog/api.py
YS-L: Good point, will check if we can simply replace the existing route.

PR #2158, opened 2024-02-26
[DEV-3062] Add redis lock to Feast registry storage update

gee-senbong:
## Description

Add redis lock to Feast registry storage update.

## Type of Change

## Checklist

### DIFF
--- a/featurebyte/feast/service/registry.py
+++ b/featurebyte/feast/service/registry.py
@@ -10,6 +10,7 @@ from pathlib import Path
 
 from bson import ObjectId
 from redis import Redis
+from redis.lock import Lock
 
 from featurebyte.feast.model.registry import FeastRegistryModel
 from featurebyte.feast.schema.registry import FeastRegistryCreate, FeastRegistryUpdate
@@ -25,6 +26,8 @@ from featurebyte.service.feature import FeatureService
 from featurebyte.service.feature_store import FeatureStoreService
 from featurebyte.service.online_store import OnlineStoreService
 from featurebyte.storage import Storage
+
+FEAST_REGISTRY_REDIS_LOCK_TIMEOUT = 120  # a maximum life for the lock in seconds
 
 
 class FeastRegistryService(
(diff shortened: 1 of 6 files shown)


featurebyte/feast/service/registry.py
```diff
@@ -223,7 +241,9 @@ async def _populate_registry(self, document: FeastRegistryModel) -> FeastRegistr
 
     async def _move_registry_to_storage(self, document: FeastRegistryModel) -> FeastRegistryModel:
         feast_registry_path = Path(f"feast_registry/{document.id}/feast_registry.pb")
-        await self.storage.put_bytes(document.registry, feast_registry_path)
+        with self.get_registry_storage_lock(timeout=FEAST_REGISTRY_REDIS_LOCK_TIMEOUT):
```
kchua78, line 244: I think we may need to get the lock earlier before getting the enabled feature lists.
Both `_get_online_enabled_feature_lists` and  `_create_or_update_feast_registry` needs to be done after acquiring the lock and completed before releasing

gee-senbong: I don't fully get the idea behind this. The main purpose here is to prevent concurrent update of the registry path. If we want to extend it to the feature list update, how about feature update?

gee-senbong: Or we delegate the feature update to the offline table Redis lock? If that is the case, i think it makes sense.

featurebyte/feast/service/registry.py
```diff
@@ -287,7 +307,8 @@ async def update_document(
 
         if original_doc.registry_path:
             # remove old registry file
-            await self.storage.delete(Path(original_doc.registry_path))
+            with self.get_registry_storage_lock(timeout=FEAST_REGISTRY_REDIS_LOCK_TIMEOUT):
+                await self.storage.delete(Path(original_doc.registry_path))
 
         document = await self._move_registry_to_storage(recreated_model)
```
YS-L, line 313: If only this part is locked, could the registry document in mongo be out of sync with the registry in the storage? This might or might not matter, but I was under the impression that the lock would cover the whole duration of `update_document` to make the update operation atomic.

YS-L (approved):

PR #2160, opened 2024-02-26
[DEV-2994] Enable Feast integration by default

gee-senbong:
## Description

This PR enable Feast integration by default.

## Type of Change

## Checklist


featurebyte/service/offline_store_feature_table_manager.py
```diff
@@ -419,10 +429,33 @@ async def _get_required_aggregate_result_tables(
...
+        fl_has_been_enabled = False
         async for feature_list_dict in self.feature_list_service.iterate_online_enabled_feature_lists_as_dict():
-            feature_lists.append(FeatureListModel(**feature_list_dict))
+            feature_list = FeatureListModel(**feature_list_dict)
+            if (
+                feature_list_to_online_disable
+                and feature_list.id == feature_list_to_online_disable.id
+            ):
+                # skip the feature list to be disabled
+                continue
```
YS-L, line 446: We need this check because now the feature list's `enabled` flag is updated at the very last step of the deployment update process, is that correct?

tests/integration/query_graph/test_online_serving.py
```diff
@@ -89,6 +90,9 @@ def features_fixture(event_table, source_type):
     return features
 
 
+@pytest.mark.parametrize(
+    "source_type", SNOWFLAKE_SPARK_DATABRICKS, indirect=True  # skip unity catalog: DEV-3109
```
YS-L, line 94: Is this an issue that rarely occurs? I think we unconditionally skip the `test / test-integration-databricks (pull_request) Skipped` suite now, so `SNOWFLAKE_SPARK_DATABRICKS` is effectively the same as `SNOWFLAKE_SPARK`.

featurebyte/models/deployment.py
```diff
@@ -58,4 +58,4 @@ class FeastIntegrationSettings(BaseSettings):
     Feast integration settings
     """
 
-    FEATUREBYTE_FEAST_INTEGRATION_ENABLED: bool = False
+    FEATUREBYTE_FEAST_INTEGRATION_ENABLED: bool = True
```
YS-L, line 61: All the integration tests including databricks_unity run with this flag enabled, is that correct?

featurebyte/service/offline_store_feature_table_manager.py
gee-senbong: Yes, the feature list's attribute is updated at later step in `DeployService`.
```

        if feast_registry:
            await self.feast_integration_service.handle_online_enabled_features(  # this calls offline store ft manager 
                features=features_offline_feature_table_to_update,
                feature_list_to_online_enable=feature_list,
                update_progress=get_ranged_progress_callback(
                    self.task_progress_updater.update_progress, 72, 98
                ),
            )
            await self._update_progress(
                99, f"Updating deployed feature list ({feature_list.name}) ..."
            )
            await self.feast_integration_service.handle_deployed_feature_list(   # feature list attr updated here
                feature_list=feature_list, online_enabled_features=features
            )
```

tests/integration/query_graph/test_online_serving.py
gee-senbong: I think this issue occurs deterministically. After we fixed other flaky tests.

featurebyte/service/offline_store_feature_table_manager.py
YS-L: I see, in that case it seems like an existing bug if I understood correctly. Good to fix it 👍

featurebyte/models/deployment.py
gee-senbong: Yes, all the tests will assume Feast integration is enabled by default unless we explicitly mock it to false.

tests/integration/query_graph/test_online_serving.py
YS-L: Oh I meant whether it's an issue that will be easily encountered by users, or more a corner case that rarely occurs in real life.

featurebyte/models/deployment.py
YS-L: Got it, wanted to confirm since based on our offline discussion the databricks unity integration tests would run with the flag disabled. Seems like that is no longer required.

YS-L (approved):

featurebyte/service/offline_store_feature_table_manager.py
gee-senbong: Previously we don't rely on the feature list attribute to check whether the Feast integration is enabled or not. Strictly speaking, it is not a bug.

tests/integration/query_graph/test_online_serving.py
gee-senbong: I suspect we need to fix it. But it seems like `Feast` doesn't expect `np.float32` type (based on my rough understanding of the exception message show in the ticket). https://featurebyte.atlassian.net/browse/DEV-3109

featurebyte/models/deployment.py
gee-senbong: Ya, it is no longer required as this is the only integration test that failed with Feast integration with unity catalog at the moment.

PR #2163, opened 2024-02-27
[DEV-3065] Add redis lock for offline store table writes

YS-L:
## Description

This adds redis lock for offline store table write operations. This is to prevent concurrent writes to the same offline store table between scheduled task and manually triggered deployment task.

## Type of Change

## Checklist

### DIFF
--- a/featurebyte/service/feature_materialize.py
+++ b/featurebyte/service/feature_materialize.py
@@ -268,32 +268,17 @@ class FeatureMaterializeService:  # pylint: disable=too-many-instance-attributes
         feature_table_model: OfflineStoreFeatureTableModel
             OfflineStoreFeatureTableModel object
         """
-        with self.get_table_update_lock(offline_store_table_name=feature_table_model.name):
-            feature_timestamp = await self._scheduled_materialize_features_offline(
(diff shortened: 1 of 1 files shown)


featurebyte/service/feature_materialize.py
```diff
@@ -44,6 +46,8 @@
 from featurebyte.service.session_manager import SessionManagerService
 from featurebyte.session.base import BaseSession
 
+OFFLINE_STORE_TABLE_REDIS_LOCK_TIMEOUT_SECONDS = 3600
```
gee-senbong, line 49: When populating entity lookup table (for item table), it can takes more than one hour (local testing takes about 2-3 hours). We may need to increase this number.

gee-senbong (approved):

featurebyte/service/feature_materialize.py
YS-L: Updated the logic to reduce the scope of lock to offline store table writes only, excluding online store materialize. With that I think the current timeout should be sufficient. The slow part is online store update materialize we should further investigate.

PR #2165, opened 2024-02-28
[DEV-3068] Move feature clusters to remote file

gee-senbong:
## Description

This PR moves feature list's feature clusters to remote storage, offline feature table's feature clusters will be fixed in another PR.

## Type of Change

## Checklist

### DIFF
--- a/featurebyte/service/feature_list.py
+++ b/featurebyte/service/feature_list.py
@@ -11,7 +11,6 @@ from pathlib import Path
 from bson import json_util
 from bson.objectid import ObjectId
 from redis import Redis
-from redis.lock import Lock
 
 from featurebyte.common.model_util import get_version
 from featurebyte.exception import DocumentError, DocumentInconsistencyError, DocumentNotFoundError
@@ -139,6 +138,7 @@ class FeatureListService(  # pylint: disable=too-many-instance-attributes
             persistent=persistent,
             catalog_id=catalog_id,
             block_modification_handler=block_modification_handler,
+            storage=storage,
             redis=redis,
         )
         self.feature_store_service = feature_store_service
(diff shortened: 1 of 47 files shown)


featurebyte/service/feature_list.py
```diff
@@ -320,6 +351,7 @@ async def create_document(self, data: FeatureListServiceCreate) -> FeatureListMo
                 "features_entity_lookup_info": entity_relationship_data.features_entity_lookup_info,
             }
         )
+        await self._move_feature_cluster_to_storage(document)
```
YS-L, line 354: Can the remaining steps below fail? If so might be better to add a step to clean up the remote file when that occurs.

tests/unit/conftest.py
```diff
@@ -128,17 +128,17 @@ def storage_fixture():
     """
     Storage object fixture
     """
-    with tempfile.TemporaryDirectory() as tempdir:
-        yield LocalStorage(base_path=tempdir)
+    with tempfile.TemporaryDirectory(suffix=f"_{ObjectId()}") as tempdir:
+        yield LocalStorage(base_path=Path(tempdir))
```
YS-L, line 132: With this, do different tests still use a shared storage making the `delete_document` step in the tests necessary?

tests/unit/service/conftest.py
```diff
@@ -620,10 +633,15 @@ async def feature_list_repeated_fixture(test_dir, feature, feature_list_service)
...
-        return feature_list
+        feature_list = None
+        try:
+            feature_list = await feature_list_service.create_document(
+                data=FeatureListServiceCreate(**payload)
+            )
+            yield feature_list
+        finally:
+            if feature_list:
+                await feature_list_service.delete_document(document_id=feature_list.id)
```
YS-L, line 644: This `create_document` and `delete_document` pattern seems common enough that a helper context manager might be useful.

YS-L: Btw is it true that all tests that create a feature list will need to remember to clean up after use? Or only for specific types of tests?

YS-L (approved):

tests/unit/conftest.py
gee-senbong: Technically, it is not necessary.

tests/unit/service/conftest.py
gee-senbong: When the storage type is not function scope, would be good to clean it up when we create the feature list using fixed feature list ID (otherwise, between tests may have coupling effect). I encountered this a couple of times when fixing the tests (maybe before i added the suffix), found that it is useful to isolate the problem by just cleaning up the feature list.

PR #2186, opened 2024-03-06
[chore] reduce / simplify queries for source table operations

kchua78:
## Description

- Avoid multiple datawarehouse calls when getting table schema
- Use simpler SQL for getting shape and preview for source tables
- Exclude featurebyte internal tables in table listing
- Exclude timeout from session cache hash
- Use CPU worker for historical features computation

## Type of Change

## Checklist

### DIFF
--- a/featurebyte/schema/worker/task/historical_feature_table.py
+++ b/featurebyte/schema/worker/task/historical_feature_table.py
@@ -8,7 +8,7 @@ from typing import Optional
 from featurebyte.enum import WorkerCommand
 from featurebyte.models.historical_feature_table import HistoricalFeatureTableModel
 from featurebyte.schema.historical_feature_table import HistoricalFeatureTableCreate
-from featurebyte.schema.worker.task.base import BaseTaskPayload
+from featurebyte.schema.worker.task.base import BaseTaskPayload, TaskType
 
 
 class HistoricalFeatureTableTaskPayload(BaseTaskPayload, HistoricalFeatureTableCreate):
@@ -18,4 +18,5 @@ class HistoricalFeatureTableTaskPayload(BaseTaskPayload, HistoricalFeatureTableC
 
     output_collection_name = HistoricalFeatureTableModel.collection_name()
     command = WorkerCommand.HISTORICAL_FEATURE_TABLE_CREATE
+    task_type = TaskType.IO_TASK
     observation_set_storage_path: Optional[str]
--- a/featurebyte/schema/worker/task/historical_feature_table.py
+++ b/featurebyte/schema/worker/task/historical_feature_table.py
@@ -18,5 +18,5 @@ class HistoricalFeatureTableTaskPayload(BaseTaskPayload, HistoricalFeatureTableC
 
     output_collection_name = HistoricalFeatureTableModel.collection_name()
     command = WorkerCommand.HISTORICAL_FEATURE_TABLE_CREATE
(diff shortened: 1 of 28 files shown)


featurebyte/schema/worker/task/historical_feature_table.py
```diff
@@ -18,4 +18,5 @@ class HistoricalFeatureTableTaskPayload(BaseTaskPayload, HistoricalFeatureTableC
 
     output_collection_name = HistoricalFeatureTableModel.collection_name()
     command = WorkerCommand.HISTORICAL_FEATURE_TABLE_CREATE
+    task_type = TaskType.IO_TASK
```
gee-senbong, line 21: Should we change this to CPU task?

featurebyte/session/base.py
```diff
@@ -496,7 +496,7 @@ async def execute_query(
         return dataframe_from_arrow_stream(data)
 
     async def execute_query_interactive(
-        self, query: str, timeout: float = INTERACTIVE_SESSION_TIMEOUT_SECONDS
+        self, query: str, timeout: float = DEFAULT_EXECUTE_QUERY_TIMEOUT_SECONDS
     ) -> pd.DataFrame | None:
         """
         Execute SQL query that is expected to run for a short time
```
YS-L, line 502: Is `DEFAULT_EXECUTE_QUERY_TIMEOUT_SECONDS` considered short enough? What is the consequence if the query ended up running for a long time when calling this method?

featurebyte/service/preview.py
```diff
@@ -295,6 +295,83 @@ async def value_counts(
...
+            for col_name in columns_specs.keys()
+            if col_name != InternalName.TABLE_ROW_INDEX
+        ]
+        return (
+            (result["row_count"].iloc[0], len(columns)),
+            has_row_index,
+            columns,
+        )
+
+    async def table_shape(self, location: TabularSource) -> FeatureStoreShape:
```
YS-L, line 322: Where is `table_shape` and `table_preview` called in the code base?

featurebyte/session/base.py
kchua78: Looks like `DEFAULT_EXECUTE_QUERY_TIMEOUT_SECONDS` is for `def execute_query()`. Will revert this. Setting the timeout too long for interactive query can cause threads to accumulate when traffic to the api service is heavy

featurebyte/service/feature_store_warehouse.py
```diff
@@ -106,6 +131,37 @@ async def list_schemas(
...
 
+    @staticmethod
+    def _is_visible_table(table_name: str, filter_featurebyte_tables: bool) -> bool:
+        if table_name.startswith("__"):
+            return False
+        if not filter_featurebyte_tables:
+            return True
+        # quick filter for materialized tables
+        if "TABLE" not in table_name:
+            return False
```
gee-senbong, line 142: Curious that 
```
        for prefix in MaterializedTableNamePrefix.visible():
            if table_name.startswith(prefix):
                return True
```
should be quite fast, right? Do we need this?

kchua78: The majority of tables are not materialized tables (most are tile tables). With this check we can reject these using a string comparison operation instead of multiple checks against each prefix

YS-L (approved):

PR #2192, opened 2024-03-07
[chore] Fix FeatureMaterializeService handling of existing but empty feature tables

YS-L:
## Description

This fixes an error that occurs during `_initialize_new_columns_offline`  when a feature table exists but is empty. This is uncommon but could occur due to empty entity universe when the feature table was created.

## Type of Change

## Checklist


featurebyte/service/feature_materialize.py
```diff
@@ -347,7 +349,7 @@ async def _initialize_new_columns_offline(
             selected_columns=selected_columns,
             use_last_materialized_timestamp=False,
         ) as materialized_features:
-            if not has_existing_table:
+            if num_rows_in_feature_table is None:
                 # Create feature table is it doesn't exist yet
                 await self._create_feature_table(
                     session,
```
gee-senbong, line 355: Will it drop the table if the table already exists?

YS-L: If the table exists, it won't be deleted here. A feature table will be deleted during online disabling when it no longer has any columns (all of its features were online disabled).

gee-senbong: In this case, will it case any issue if we recreate an empty table (since line 354 will create a new table)?

YS-L: Line 354 creating an empty table is fine. The issue is when the table already exists: previously we assumed that we can always call `_get_last_feature_timestamp` to get a non-null value, and the merge query was invalid when the table is empty.

gee-senbong: Now i got it, actually the condition is the same (checking `num_rows_in_feature_table is None` is same as the table does not exist).

PR #2218, opened 2024-03-17
[chore] Handle schema inconsistency between query fetch batches

kchua78:
## Description

- Use cursor descriptions as source of truth for query return schema
- Standardize data format for pyarrow record batches for different warehouses

## Type of Change

## Checklist

### DIFF
--- a/docker/docker-compose.yml
+++ b/docker/docker-compose.yml
@@ -49,7 +49,7 @@ services:
         --hiveconf hive.server2.thrift.bind.host=0.0.0.0 \
         --hiveconf derby.system.home=/opt/spark/data/derby \
         --master=local[2] \
-        --conf spark.driver.memory=2G \
+        --conf spark.driver.memory=3G \
         --conf spark.sql.extensions=io.delta.sql.DeltaSparkSessionExtension \
         --conf spark.sql.catalog.spark_catalog=org.apache.spark.sql.delta.catalog.DeltaCatalog \
         --conf spark.sql.catalogImplementation=hive  \
--- a/featurebyte/common/utils.py
+++ b/featurebyte/common/utils.py
@@ -1,5 +1,5 @@
 """
-Utility functions for API Objects
+Common utility functions
 """
 from __future__ import annotations
 
@@ -15,16 +15,12 @@ from contextlib import contextmanager
 from datetime import datetime
 from decimal import Decimal
 from importlib import metadata as importlib_metadata
-from pathlib import Path
 
 import numpy as np
 import pandas as pd
 import pyarrow as pa
-import pyarrow.parquet as pq
-from alive_progress import alive_bar
 from dateutil import parser
 
-from featurebyte.common.env_util import get_alive_bar_additional_params
 from featurebyte.enum import DBVarType, InternalName
 
 ARROW_METADATA_DB_VAR_TYPE = b"db_var_type"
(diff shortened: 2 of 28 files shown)


featurebyte/common/utils.py
```diff
@@ -119,22 +120,54 @@ def dataframe_to_arrow_bytes(dataframe: pd.DataFrame) -> bytes:
...
+        arrow_table = reader.read_all()
+    else:
+        batches = []
+        batches.append(reader.read_next_batch())
+        try:
+            with alive_bar(
+                total=num_rows,
+                title="Downloading table",
+                **get_alive_bar_additional_params(),
+            ) as progress_bar:
```
YS-L, line 150: Is this `dataframe_from_arrow_stream` function only used by API objects but not in the services? If so would be good to move it to under the `api` directory to make that clear. I think using `alive_bar` in backend service code might cause issues due to the use of threading etc.

featurebyte/session/snowflake.py
```diff
@@ -186,51 +230,25 @@ def fetch_query_result_impl(self, cursor: Any) -> pd.DataFrame | None:
...
-            if counter == 0:
-                # Arrow batch is empty, need to return empty table with schema
-                table = cursor.get_result_batches()[0].to_arrow()
-                yield pa_table_to_record_batches(table)[0]
-        except NotSupportedError:
+            # return empty table to ensure correct schema is returned
+            yield pa.record_batch(
+                pd.DataFrame(columns=[field.name for field in schema]), schema=schema
+            )
+        except (NotSupportedError, AttributeError):
```
YS-L, line 245: Where is AttributeError expected to be raised? Would be good to leave a comment on that since it's not obvious. If the exception type is something we can control, it would be nice to raise a more specific exception here since AttributeError is rather generic.

tests/integration/api/test_event_view_operations.py
```diff
@@ -147,13 +147,12 @@ def test_event_view_ops(event_view, transaction_data_upper_case, source_type):
     columns = [
         col for col in output.columns if not col.startswith("str_") and not col.startswith("dt_")
     ]
-    if source_type == "spark":
+
+    # only snowflake supports timezone in datetime columns
+    if source_type not in ["snowflake"]:
```
YS-L, line 152: Is this a change in behaviour for databricks?

docker/docker-compose.yml
```diff
@@ -49,7 +49,7 @@ services:
         --hiveconf hive.server2.thrift.bind.host=0.0.0.0 \
         --hiveconf derby.system.home=/opt/spark/data/derby \
         --master=local[2] \
-        --conf spark.driver.memory=2G \
+        --conf spark.driver.memory=3G \
```
gee-senbong, line 52: Is it intentional to increase the RAM amount here? If the RAM usage goes up, we may have the problem of starting up the app as reported previously.

featurebyte/common/utils.py
gee-senbong: Ya, i also encountered `dictionary changed size during iteration` before when using `alive_bar`. Would be good to move it to `api` specific directory to prevent people use it in backend service.

docker/docker-compose.yml
kchua78: This compose file only applies specifically to the test spark instance used for integration testing.

gee-senbong: Got it, I thought this is `featurebyte-app`. 😅

featurebyte/common/utils.py
kchua78: The function is used in both places currently but alive bar is only used from the api (by passing in num_rows) to display download progress. I can refactor this into 2 different functions and move one under api

featurebyte/session/spark.py
```diff
@@ -241,150 +236,53 @@ def delete_path_from_storage(self, remote_path: str) -> None:
...
         # fetch results in batches
-        schema = pa.schema(
-            {metadata[0]: self._get_pyarrow_type(metadata[1]) for metadata in cursor.description}
-        )
+        schema = self._get_schema_from_cursor(cursor)
         while True:
             try:
-                record_batch = self._read_batch(cursor, schema)
-            except TypeError:
+                records = cursor.fetchmany(1000)
```
gee-senbong, line 278: Would be good to add a constant variable for this hard-coded value.

gee-senbong (approved):

tests/integration/api/test_event_view_operations.py
kchua78: This is expected since we use cursor description to determine pyarrow type now instead of using the type detected by pandas conversion

YS-L (approved):

PR #2251, opened 2024-03-28
[chore] Cache `FeatureList`'s & `FeastRegistry`'s remote attribute 

gee-senbong:
## Description

This PR adds caching mechanism for `FeatureList` & `FeastRegistry` to improve runtime when retrieving the same object more than 1 time.

## Type of Change

## Checklist


featurebyte/storage/base.py
```diff
@@ -67,22 +72,58 @@ async def try_delete_if_exists(self, remote_path: Path) -> None:
...
+        # cache miss
+        await self._get(remote_path, local_path)
+        try:
+            async with aiofiles.open(local_path, "rb") as file_obj:
+                value = await file_obj.read()
+
+            async with self._cache_lock:
+                self._cache.setdefault(cache_key, value)
+        except ValueError:
+            pass  # value too large
```
YS-L, line 114: Is this a limit that can be configured?

gee-senbong: This is based on the implementation in `cachedtools` library:
https://github.com/tkem/cachetools/blob/master/src/cachetools/__init__.py#L781-L844

featurebyte/storage/base.py
```diff
@@ -67,22 +72,58 @@ async def try_delete_if_exists(self, remote_path: Path) -> None:
...
 
         Parameters
         ----------
         remote_path: Path
             Path of remote file to be downloaded
         local_path: Path
             Path to stored downloaded file
+        cache_key: Any
+            Cache key for storing downloaded file (if provided, the result will be cached).
+            Please do not use cache_key for large files.
```
YS-L, line 87: Typically we want to use caching for files that can be large such as feature clusters and registry, is that correct?

gee-senbong: Since we want to load the file into memory, we don't want to load like 1GB data into memory which can cause other issue. Feature clusters & registry is more like in several 10MB range, which is much smaller than 1GB.

featurebyte/storage/base.py
gee-senbong: This is controlled by the cache size (https://github.com/tkem/cachetools/blob/master/src/cachetools/__init__.py#L72-L86).

featurebyte/storage/base.py
YS-L: What do you think about caching the file on disk instead? Then the lru cache can track the cached filename and doesn't need to keep file content in memory.

gee-senbong: In this case, maybe we put it under `.featurebyte/cache` directory?

gee-senbong: I think i have some ideas, let's me revise this & ping you once it is revised.

YS-L: Either seems fine to me. I think we should check whether the file still exists upon cache hit.

YS-L (approved):

PR #2278, opened 2024-04-05
[chore] Support concurrency with asyncio gather

kchua78:
## Description

Experimental PR to add concurrency limit to parallel query execution using asyncio

## Type of Change

## Checklist

### DIFF
--- a/featurebyte/session/session_helper.py
+++ b/featurebyte/session/session_helper.py
@@ -6,6 +6,8 @@ from __future__ import annotations
 
 from typing import Any, Callable, Coroutine, List, Optional, Union
 
+import os
+
 import pandas as pd
 from sqlglot import expressions
 from sqlglot.expressions import Expression
@@ -21,7 +23,7 @@ from featurebyte.utils.async_helper import asyncio_gather
 logger = get_logger(__name__)
 
 
-MAX_QUERY_CONCURRENCY = 10
+MAX_QUERY_CONCURRENCY = int(os.getenv("MAX_QUERY_CONCURRENCY", "10"))
 
 
 def _to_query_str(query: Union[str, Expression], source_type: SourceType) -> str:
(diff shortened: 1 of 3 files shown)


vikua (approved):

featurebyte/session/session_helper.py
```diff
@@ -88,16 +92,11 @@ async def run_coroutines(coroutines: List[Coroutine[Any, Any, Any]]) -> List[Any
...
-        return cast(List[Any], await asyncio.gather(*tasks))
+        return await future
     except Exception:
-        logger.error(
-            "Canceling all other tasks because at least one task failed",
-            exc_info=True,
-        )
-        for task in tasks:
-            task.cancel()
+        future.cancel()
```
YS-L, line 99: The tasks are already cancelled on error in `asyncio_gather`. Is this cancel still required?

featurebyte/utils/async_helper.py
```diff
@@ -0,0 +1,55 @@
...
+                return
+            task = asyncio.create_task(coro)
+            tasks.append(task)
+            try:
+                return await task
+            except Exception:
+                failed = True
+                # Cancel all tasks on failure
+                for task in tasks:
+                    task.cancel()
```
YS-L, line 52: This loop over tasks is done within each coroutine, is that intended? It seems each task can potentially be cancelled more than once.

featurebyte/session/session_helper.py
kchua78: These tasks are the ones created for the coroutines returned by the wrapper function `_coro` and are different from the tasks in `asyncio_gather` which are created by `asyncio.create_task(coro)`

kchua78: I think you are right this is not necessary since the wrapper tasks should all terminate when the task they created are canceled. Will remove this

featurebyte/utils/async_helper.py
kchua78: This is intended as the caller of the `asyncio_gather` function has no access to the tasks created by the `_coro` wrapper functions. It seems that calling cancel on the same task multiple times does raise any error but I can add a flag to avoid the multiple calls

PR #2287, opened 2024-04-11
[DEV-3230] Create precomputed lookup feature table records 

YS-L:
## Description

This updates offline store feature table manager to start creating a new kind of offline store feature table to support precomputed parent entity lookup in feature tables. Materialization of such tables will be added in an upcoming PR.

## Type of Change

## Checklist

### DIFF
--- a/featurebyte/models/offline_store_feature_table.py
+++ b/featurebyte/models/offline_store_feature_table.py
@@ -43,6 +43,17 @@ class OnlineStoreLastMaterializedAt(BaseModel):
     value: datetime
 
 
+class PrecomputedLookupFeatureTableInfo(BaseModel):
+    """
+    Metadata for a feature table that is derived from a source feature table in order to support
+    precomputed lookup using a related entity
+    """
+
+    lookup_steps: List[EntityRelationshipInfo]
+    feature_list_ids: List[PydanticObjectId] = Field(default_factory=list)
+    source_feature_table_id: Optional[PydanticObjectId]
+
+
 class OfflineStoreFeatureTableModel(FeatureByteCatalogBaseDocumentModel):
     """
     OfflineStoreFeatureTable class
@@ -68,6 +79,7 @@ class OfflineStoreFeatureTableModel(FeatureByteCatalogBaseDocumentModel):
     output_dtypes: List[DBVarType]
     internal_entity_universe: Optional[Dict[str, Any]] = Field(alias="entity_universe")
     entity_lookup_info: Optional[EntityRelationshipInfo]
+    precomputed_lookup_feature_table_info: Optional[PrecomputedLookupFeatureTableInfo]
     feature_store_id: Optional[PydanticObjectId] = Field(default=None)
 
     @root_validator
@@ -108,6 +120,11 @@ class OfflineStoreFeatureTableModel(FeatureByteCatalogBaseDocumentModel):
         entity_lookup_info = None
         if self.entity_lookup_info:
             entity_lookup_info = self.entity_lookup_info.dict(by_alias=True)
+        precomputed_lookup_feature_table_info = None
+        if self.precomputed_lookup_feature_table_info:
+            precomputed_lookup_feature_table_info = self.precomputed_lookup_feature_table_info.dict(
+                by_alias=True
+            )
         return {
             "catalog_id": self.catalog_id,
             "primary_entity_ids": self.primary_entity_ids,
@@ -117,6 +134,7 @@ class OfflineStoreFeatureTableModel(FeatureByteCatalogBaseDocumentModel):
             ),
             "has_ttl": self.has_ttl,
             "entity_lookup_info": entity_lookup_info,
+            "precomputed_lookup_feature_table_info": precomputed_lookup_feature_table_info,
         }
 
     @property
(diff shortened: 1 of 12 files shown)


featurebyte/models/precomputed_lookup_feature_table.py
```diff
@@ -0,0 +1,291 @@
...
+                                entity_lookup_steps_mapping=entity_lookup_steps_mapping,
+                                feature_store=feature_store_model,
+                            )
+                        )
+                    ),
+                    precomputed_lookup_feature_table_info=PrecomputedLookupFeatureTableInfo(
+                        lookup_steps=lookup_steps,
+                        source_feature_table_id=feature_table_id,
+                    ),
+                    has_ttl=False,
```
gee-senbong, line 176: Is this always False?

featurebyte/service/feature_materialize.py
```diff
@@ -542,10 +542,12 @@ async def _get_feast_feature_store(self) -> Optional[FeastFeatureStore]:
         return await self.feast_feature_store_service.get_feast_feature_store(feast_registry)
 
     async def _get_session(self, feature_table_model: OfflineStoreFeatureTableModel) -> BaseSession:
-        assert feature_table_model.feature_cluster is not None, "Missing feature cluster"
-        feature_store = await self.feature_store_service.get_document(
-            document_id=feature_table_model.feature_cluster.feature_store_id
-        )
+        if feature_table_model.feature_cluster is not None:
+            feature_store_id = feature_table_model.feature_store_id
```
kchua78, line 546: should this be `feature_table_model.feature_cluster.feature_store_id`?

featurebyte/models/offline_store_feature_table.py
```diff
@@ -43,6 +43,17 @@ class OnlineStoreLastMaterializedAt(BaseModel):
...
 
 
+class PrecomputedLookupFeatureTableInfo(BaseModel):
+    """
+    Metadata for a feature table that is derived from a source feature table in order to support
+    precomputed lookup using a related entity
+    """
+
+    lookup_steps: List[EntityRelationshipInfo]
+    feature_list_ids: List[PydanticObjectId] = Field(default_factory=list)
```
kchua78, line 53: Do we need the feature list ids for materialization of these derived feature tables?

featurebyte/models/precomputed_lookup_feature_table.py
YS-L: I think this should be set the same as the source feature table's ttl. Will fix.

featurebyte/service/feature_materialize.py
YS-L: Yes, though there is a root validator that sets `feature_store_id` based on the feature cluster:

https://github.com/featurebyte/featurebyte/blob/ae1f0e8e89abf071b495db3405f79f4425d71d94/featurebyte/models/offline_store_feature_table.py#L85-L102

Will change back to `feature_table_model.feature_cluster.feature_store_id` to be safe and not break backward compatibility.

featurebyte/models/offline_store_feature_table.py
YS-L: Feature list ids are not needed for materialization, but needed for feast registry construction (not included in this PR).

kchua78 (approved):

featurebyte/models/offline_store_feature_table.py
```diff
@@ -68,6 +79,7 @@ class OfflineStoreFeatureTableModel(FeatureByteCatalogBaseDocumentModel):
     output_dtypes: List[DBVarType]
     internal_entity_universe: Optional[Dict[str, Any]] = Field(alias="entity_universe")
     entity_lookup_info: Optional[EntityRelationshipInfo]
+    precomputed_lookup_feature_table_info: Optional[PrecomputedLookupFeatureTableInfo]
```
gee-senbong, line 82: To check my understanding, when this attribute it not empty, it means this table is derived from other existing offline feature table, right?

featurebyte/models/precomputed_lookup_feature_table.py
```diff
@@ -0,0 +1,294 @@
...
+
+    Returns
+    -------
+    str
+    """
+    serving_names_suffix = OfflineStoreFeatureTableModel.get_serving_names_for_table_name(
+        serving_names
+    )
+    unique_identifier = get_lookup_steps_unique_identifier(lookup_steps)
+    return f"{source_feature_table_name}_via_{serving_names_suffix}_{unique_identifier}"
```
gee-senbong, line 83: Max table name length for MySQL is 64 (for DataBricks is 255). Do we need to worry about exceeding table name limit here?

featurebyte/models/precomputed_lookup_feature_table.py
```diff
@@ -0,0 +1,294 @@
...
+        combined_lookup_steps: List[EntityRelationshipInfo] = []
+        if feature_list.features_entity_lookup_info:
+            for info in feature_list.features_entity_lookup_info:
+                for step in info.join_steps:
+                    if step not in combined_lookup_steps:
+                        combined_lookup_steps.extend(info.join_steps)
+        feature_lists_relationships_info[feature_list.id] = (
+            feature_list.relationships_info or []
+        ) + combined_lookup_steps
+    return feature_lists_relationships_info
```
gee-senbong, line 294: nit: We may consider to move parts of the logic to extract feature list join steps into feature list model later.

featurebyte/service/offline_store_feature_table.py
```diff
@@ -250,3 +252,27 @@ async def update_online_last_materialized_at(
...
+        -------
+        AsyncIterator[OfflineStoreFeatureTableModel]
+            List query output
+        """
+        async for doc in self.list_documents_iterator(
+            query_filter={
+                "precomputed_lookup_feature_table_info.source_feature_table_id": source_feature_table_id
+            }
+        ):
+            yield doc
```
gee-senbong, line 278: Since we skip building precompute feature table for those feature lists do not require additional lookups, Just double whether we don't have to include source feature table here.

featurebyte/service/offline_store_feature_table_manager.py
```diff
@@ -482,6 +530,84 @@ async def _create_or_update_feast_registry(
...
+            query_filter={"precomputed_lookup_feature_table_info": None},
+            projection={"_id": 1, "feature_ids": 1},
+        ):
+            feature_ids = feature_table_dict["feature_ids"]
+            await self._update_precomputed_lookup_feature_tables(
+                feature_table_dict["_id"],
+                feature_ids,
+                feature_lists,
+                feature_store_model,
+            )
```
gee-senbong, line 552: The main reason of doing this is because of new relationship that may not be covered in existing pre-computed lookup feature tables, right?

gee-senbong: It should be a new feature list, right? If existing feature list, the relationship should be freeze during feature list construction.

gee-senbong (approved):

featurebyte/models/offline_store_feature_table.py
YS-L: Yes, that is correct.

featurebyte/models/precomputed_lookup_feature_table.py
YS-L: That might occur since `source_feature_table_name` can reach that limit, though there should be plenty of room for databricks. I suggest we revisit the limits again later to account for the new suffixes.

featurebyte/service/offline_store_feature_table.py
YS-L: It is intended for this method to return only the additional lookup tables, or return an empty list for source tables that do not require additional lookups.

featurebyte/service/offline_store_feature_table_manager.py
YS-L: Yes, this is to handle the new feature list that is being online enabled.

PR #2312, opened 2024-04-23
[DEV-3292] Refactor `FeatureJobSetting` parameters

gee-senbong:
## Description

This PR refactor `FeatureJobSetting` attribute to match new naming convention.

## Type of Change

## Checklist


featurebyte/query_graph/node/generic.py
```diff
@@ -685,14 +686,24 @@ class GroupByNodeParameters(BaseGroupbyParameters):
 
     windows: List[Optional[str]]
     timestamp: InColumnStr
-    blind_spot: int
-    time_modulo_frequency: int
-    frequency: int
     names: List[OutColumnStr]
+    feature_job_setting: FeatureJobSetting
```
gee-senbong, line 690: Refactor this part by adding a nested `FeatureJobSetting` object. By doing this, it makes us easier to extend for other feature job setting format. For example:
```
feature_job_setting: Union[FeatureJobSetting, CronJobFeatureJobSetting]
```

featurebyte/query_graph/model/feature_job_setting.py
```diff
@@ -73,38 +78,58 @@ def validate_setting_parameters(cls, values: Dict[str, Any]) -> Dict[str, Any]:
...
-            time_modulo_frequency=values["time_modulo_frequency"],
+            period=values["period"],
+            offset=values["offset"],
             blind_spot=values["blind_spot"],
         )
+
+        # normalize and convert to seconds
+        values["period"] = f'{parse_duration_string(values["period"])}s'
+        values["offset"] = f'{parse_duration_string(values["offset"])}s'
+        values["blind_spot"] = f'{parse_duration_string(values["blind_spot"])}s'
```
YS-L, line 107: Should `execution_buffer` be normalised as well though it's not yet supported?

featurebyte/schema/feature_job_setting_analysis.py
```diff
@@ -112,11 +112,33 @@ class FeatureJobSetting(FeatureByteBaseModel):
     Feature Job Setting
     """
 
-    frequency: int
-    job_time_modulo_frequency: int
```
YS-L, line 116: Will this break the UI? Wondering if it's possible to make this backward compatible for a smoother transition.

gee-senbong: There are list of items that we need to update the UI. This is not the only place that requires changes in UI. 
If we want to do that, we may need to keep both old & new parameters appear at the response, and at the same time, our migration should only keep the new feature job setting format. That means we have the `FeatureJobSetting` behave differently when it is at the response versus the format used in persistence. It requires more changes & updates.

YS-L: Yes, I was wondering whether it's possible to keep both old and new parameters in the response without too much work. Something like

```python
class FeatureJobSetting(FeatureByteBaseModel):
    """
    Feature Job Setting
    """

    period: int
    offset: int
    ...
    frequency: int = Field(exclude=True)
    job_time_modulo_frequency: int = Field(exclude=True)

    @root_validator(pre=True)
    @classmethod
    def _handle_backward_compatibility(cls, values: Dict[str, Any]) -> Dict[str, Any]:
        # set period / offset using frequency / job_time_modulo_frequency if provided
        ...
        # set frequency / job_time_modulo_frequency using period / offset if provided
        ...
```

and backend will still always use the new parameters `period` and `offset` and ignore the old parameters. Will this work and eliminate UI breakage?

gee-senbong: If the `exclude` is set to `True`, `feature_job_setting.dict()` will exclude those parameters too. That means response will not have these parameters set.

gee-senbong: Inside the fastapi response preparation:
https://github.com/tiangolo/fastapi/blob/0.96.0/fastapi/routing.py#L79-L84

We also use `dict` when we serialize the model into dictionary format before writing to mongo persistent.

gee-senbong: Unless we want to overwrite pydantic `BaseModel.dict` behaviour so that it can behave differently for response preparation & persistent input preparation.

YS-L: I see. How about using a different schema for read response? Then it is decoupled from what is stored in persistent.

gee-senbong: Ya, i think this is possible for those response where the feature job setting is not deeply nested inside the schema (for example feature's graph).

YS-L: I think it would be nice if we can avoid breaking UI if the effort is not too big, but otherwise I'm also fine with the current approach.

YS-L (approved):

featurebyte/schema/feature_job_setting_analysis.py
gee-senbong: Only the feature's graph level that part is difficult (skipping this deeply nested fjs inside the graph should be fine). Others should be Ok. Will update this PR & tests it on the featurebyte-app first. 🙏

.changelog/DEV-3292.yaml
```diff
@@ -0,0 +1,34 @@
+# This template file is used to generate changelog entries on release
+# Check the generated entry in your PR with the task command
+
+# To view the generated changelog, run the following command:
+# task changelog-pr
+
+# --- TEMPLATE --- #
+# One of 'breaking', 'deprecation', 'enhancement', 'bug_fix'
+change_type: enhancement
```
kchua78, line 9: Should we use breaking? Since old SDK will not work any more

YS-L (approved):

YS-L:
Would be good to add a breaking changes label to notify slack channels.

PR #2356, opened 2024-05-15
[DEV-2983] Prevent user to create UDF feature that is not deployable

gee-senbong:
## Description

This PR revises UDF input validation logic to prevent user to create UDF features that requires on-demand-function during deployment.

## Type of Change

## Checklist

### DIFF
--- a/featurebyte/models/feature.py
+++ b/featurebyte/models/feature.py
@@ -609,6 +609,28 @@ class FeatureModel(BaseFeatureModel):
 
         return values
 
+    @property
+    def used_request_column(self) -> bool:
+        """
+        Returns whether the Feature object uses request column(s) in the computation.
+
+        Returns
+        -------
+        bool
+        """
+        return self.graph.has_node_type(target_node=self.node, node_type=NodeType.REQUEST_COLUMN)
+
+    @property
+    def used_user_defined_function(self) -> bool:
+        """
+        Returns whether the Feature object uses user defined function(s) in the computation.
+
+        Returns
+        -------
+        bool
+        """
+        return self.graph.has_node_type(target_node=self.node, node_type=NodeType.GENERIC_FUNCTION)
+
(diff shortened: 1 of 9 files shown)


featurebyte/models/feature.py
```diff
@@ -609,6 +609,28 @@ def _add_tile_derived_attributes(cls, values: dict[str, Any]) -> dict[str, Any]:
...
+    @property
+    def used_request_column(self) -> bool:
+        """
+        Returns whether the Feature object uses request column(s) in the computation.
+
+        Returns
+        -------
+        bool
+        """
+        return self.graph.has_node_type(target_node=self.node, node_type=NodeType.REQUEST_COLUMN)
```
YS-L, line 621: Does it matter whether the `self.graph` here is pruned or not? For unsaved features, the graph here is not yet pruned right?

gee-senbong: For unsaved features, the graph is not pruned. Is it possible to assign feature with request column to a view? If yes, the output of this property before & after pruned could be different.

YS-L: I think that should be unlikely.

YS-L (approved):

featurebyte/models/feature.py
gee-senbong: Another way is to use operation structure to determine traversed nodes, maybe it is more robust to use that.

YS-L: Good point, would be good to use that.

gee-senbong: Based on our offline discussion, we should block the case where users can add request column related feature to a view through `add_feature`. If we allow that happens, that means user could aggregate on the request column derived feature (that means we need to compute the aggregation on-demand as it is impossible to compute tiles on these types of features). Since it is not possible to add request column derived feature to an view, current implementation should cover all cases without using operation structure.

PR #2460, opened 2024-07-15
[chore] Changing linter to `ruff`

bushwhackr:
## Description

+ Deprecating pylint, black, flake8 for ruff
  + We are still using black for linting generated sdk code
+ Using ruff as the new linter
+ Migrating linting steps to new Taskfile.lint.yaml
  + `task lint` remains the same
  + `task format` => `task lint:format`

```
# Before
task lint -f  749.89s user 20.92s system 525% cpu 2:26.80 total
# After
task lint -f  379.96s user 7.43s system 434% cpu 1:29.19 total
```

## Type of Change

Label this pull request to place it under the correct category in Release Notes:

|        **Pull Request Label**         | **Category in Release Notes** |
|:-------------------------------------:|:-----------------------------:|
|       `enhancement`, `feature`        |          🚀 Features          |
| `bug`, `refactoring`, `bugfix`, `fix` |    🔧 Fixes & Refactoring     |
|       `build`, `ci`, `testing`        |    📦 Build System & CI/CD    |
|              `breaking`               |      💥 Breaking Changes      |
|            `documentation`            |       📝 Documentation        |
|            `dependencies`             |    ⬆️ Dependencies updates    |

## Checklist

- [ ] I have read the [`CODE_OF_CONDUCT.md`](https://github.com/featurebyte/featurebyte/blob/main/CODE_OF_CONDUCT.md) and [`CONTRIBUTING.md`](https://github.com/featurebyte/featurebyte/blob/main/CONTRIBUTING.md) guides.
- [ ] I have written tests for the changes made.
- [ ] I have written docstrings in [NumpyDoc format](https://numpydoc.readthedocs.io/en/latest/format.html#docstring-standard)
- [ ] I have labeled my Pull Request correctly

PR #2462, opened 2024-07-15
Revert "[chore] Changing linter to `ruff`"

bushwhackr:
Reverts featurebyte/featurebyte#2460

Reverting until the pydantic PR is in


bushwhackr:
The merged PR changes too much code. I'll redo when the bigger pydantic changes are in

PR #2533, opened 2024-08-27
[chore] Revise task executor

gee-senbong:
## Description

This PR revises task executor to conditionally provide last update.

## Type of Change

## Checklist

### DIFF
--- a/featurebyte/worker/task_executor.py
+++ b/featurebyte/worker/task_executor.py
@@ -193,7 +193,18 @@ class TaskExecutor:
                 )
 
             # Send final progress to indicate task is completed
-            await self.task_progress_updater.update_progress(percent=100)
+            task = await self.task_manager.get_task(task_id=str(self.task_id))
+            skip_last_progress_update = False
+            if (
+                task is not None
+                and task.progress is not None
+                and task.progress.get("percent") == 100
+            ):
+                # skip the last progress update if the progress is already 100%
+                skip_last_progress_update = True
+
+            if not skip_last_progress_update:
+                await self.task_progress_updater.update_progress(percent=100)
         except TaskRevokeExceptions as exc:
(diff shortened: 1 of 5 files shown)


featurebyte/worker/task_executor.py
```diff
@@ -193,7 +193,18 @@ async def execute(self) -> Any:
...
+            if (
+                task is not None
+                and task.progress is not None
+                and task.progress.get("percent") == 100
+            ):
+                # skip the last progress update if the progress is already 100%
+                skip_last_progress_update = True
+
+            if not skip_last_progress_update:
+                await self.task_progress_updater.update_progress(percent=100)
```
gee-senbong, line 207: Current sub-tasks rely on progress's metadata field to pass the information back to parent task. If the task's progress get overridden, the metadata value will be lost.

kchua78: Would it make sense to update the `update_progress` logic to retain existing message if the `message` parameter is not provided.

gee-senbong: Do you mean revise this fix to keep the existing metadata & message & only update percentage?

kchua78: Would it address the issue that you are trying to fix? Does it make sense for a more systematic fix to similar problems? If not we can keep the current approach

gee-senbong: The task will automatically update the percentage to 100% and override message & task metadata. If we want to change `update_progress` method, we need to do something like keep the last value if it is not provided. Just not sure whether there is any case where we want to set the last update message/metadata to be `None`.

kchua78: Indeed. That will require an additional parameter. Let's keep the current approach

kchua78 (approved):

PR #2577, opened 2024-09-13
[DEV-3727] Add support for sampling by time range during observation table registration

kchua78:
## Description

Add support for sampling by time range during registration of observation tables from source table or view.

## Type of Change

## Checklist

### DIFF
--- a/featurebyte/models/request_input.py
+++ b/featurebyte/models/request_input.py
@@ -183,12 +183,11 @@ class BaseRequestInput(FeatureByteBaseModel):
                 )
             )
         if time_range_conditions:
-            if self.columns is None and self.columns_rename_mapping is None:
-                # add a select * in case the expression already has a where clause
-                query_expr = expressions.Select(expressions=[expressions.Star()]).from_(
-                    query_expr.subquery()
-                )
-            query_expr = query_expr.where(expressions.And(expressions=time_range_conditions))
+            query_expr = (
+                expressions.Select(expressions=[expressions.Star()])
+                .from_(query_expr.subquery())
+                .where(expressions.And(expressions=time_range_conditions))
+            )
 
         if sample_rows is not None:
             num_rows = await self.get_row_count(session=session, query_expr=query_expr)
--- a/featurebyte/models/request_input.py
+++ b/featurebyte/models/request_input.py
@@ -188,10 +188,7 @@ class BaseRequestInput(FeatureByteBaseModel):
                 query_expr = expressions.Select(expressions=[expressions.Star()]).from_(
                     query_expr.subquery()
                 )
-            if len(time_range_conditions) == 1:
-                query_expr = query_expr.where(time_range_conditions[0])
-            else:
-                query_expr = query_expr.where(expressions.And(expressions=time_range_conditions))
+            query_expr = query_expr.where(expressions.And(expressions=time_range_conditions))
 
         if sample_rows is not None:
             num_rows = await self.get_row_count(session=session, query_expr=query_expr)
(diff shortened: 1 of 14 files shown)


featurebyte/models/request_input.py
```diff
@@ -148,9 +157,42 @@ async def materialize(
...
+                    ),
+                    expression=make_literal_value(
+                        sample_to_timestamp.isoformat(), cast_as_timestamp=True
+                    ),
+                )
+            )
+        if time_range_conditions:
+            if self.columns is None and self.columns_rename_mapping is None:
+                # add a select * in case the expression already has a where clause
+                query_expr = expressions.Select(expressions=[expressions.Star()]).from_(query_expr)
```
YS-L, line 188: Does `query_expr` need to be converted into a subquery first before using in `from_`?

kchua78: yes indeed. Will update and add a test

featurebyte/models/request_input.py
```diff
@@ -148,9 +157,44 @@ async def materialize(
...
+                    ),
+                )
+            )
+        if time_range_conditions:
+            if self.columns is None and self.columns_rename_mapping is None:
+                # add a select * in case the expression already has a where clause
+                query_expr = expressions.Select(expressions=[expressions.Star()]).from_(
+                    query_expr.subquery()
+                )
+            if len(time_range_conditions) == 1:
```
YS-L, line 191: Is this condition necessary? My impression is that `expressions.And` should work even if `time_range_conditions` has a single item.

featurebyte/models/request_input.py
```diff
@@ -148,9 +157,44 @@ async def materialize(
...
             query_expr = select_and_rename_columns(query_expr, columns, self.columns_rename_mapping)
 
+        # apply time range filter if sample_from_timestamp and sample_to_timestamp are provided
+        time_range_conditions: list[expressions.Expression] = []
+        adapter = get_sql_adapter(session.get_source_info())
+        if sample_from_timestamp is not None:
+            time_range_conditions.append(
+                expressions.GTE(
+                    this=adapter.normalize_timestamp_before_comparison(
+                        quoted_identifier(SpecialColumnName.POINT_IN_TIME)
```
YS-L, line 167: Can we assume that `POINT_IN_TIME` always exist in `query_expr`?

YS-L (approved):
Can consider adding an integration test case for this.

featurebyte/models/request_input.py
kchua78: I think `POINT_IN_TIME` is a required column for any observation table after column renaming is complete. Do you see any case that that is not the case?

featurebyte/models/request_input.py
kchua78: Looks like it's not. Will remove this

featurebyte/models/request_input.py
YS-L: I don't see any other case since this is only used for materializing observation tables so a valid one should have `POINT_IN_TIME`. Just need to make sure that the validation is performed before `query_expr` is executed (I think that is the case).

kchua78: The validation for columns are done at the time of payload creation so it should be before this is called in the worker task

featurebyte/query_graph/sql/common.py
```diff
@@ -57,6 +57,8 @@ def quoted_identifier(column_name: str) -> Expression:
     -------
     Expression
     """
+    if column_name == "*":
+        return expressions.Star()
```
YS-L, line 61: Is this still needed? I think it's better to not have this logic as it most likely means this function has been misused.

kchua78: Should we raise an error in that case?

YS-L: Yes I think we can raise an error.

PR #2591, opened 2024-09-19
[DEV-3735] Enforce application-wide concurrency limit on long running SQL execution

kchua78:
## Description

Enforce application-wide concurrency limit on long running SQL execution to avoid overwhelming the feature store.

## Type of Change

## Checklist

### DIFF
--- a/featurebyte/service/feature_table_cache.py
+++ b/featurebyte/service/feature_table_cache.py
@@ -240,6 +240,7 @@ class FeatureTableCacheService:
         if is_target:
             return await get_target(
                 session=db_session,
+                redis=self.tile_cache_service.tile_manager_service.redis,
                 graph=graph,
                 nodes=nodes_only,
                 observation_set=observation_table,
@@ -480,6 +481,7 @@ class FeatureTableCacheService:
                         column_names=[InternalName.TABLE_ROW_INDEX.value] + request_column_names,
                     ),
                     exists=True,
+                    retry=True,
                 )
             historical_features_metrics = await self._update_table(
                 feature_store=feature_store,
--- a/featurebyte/service/feature_table_cache.py
+++ b/featurebyte/service/feature_table_cache.py
@@ -2,6 +2,8 @@
 Module for managing physical feature table cache as well as metadata storage.
 """
 
+from __future__ import annotations
+
 import time
 from typing import Any, Callable, Coroutine, Dict, List, Optional, Tuple, cast
 
--- a/featurebyte/service/feature_table_cache.py
+++ b/featurebyte/service/feature_table_cache.py
@@ -7,6 +7,7 @@ from typing import Any, Callable, Coroutine, Dict, List, Optional, Tuple, cast
 
 import pandas as pd
 from bson import ObjectId
+from redis import Redis
 from sqlglot import expressions
 
 from featurebyte.common.progress import get_ranged_progress_callback
@@ -60,6 +61,7 @@ class FeatureTableCacheService:
         entity_validation_service: EntityValidationService,
         tile_cache_service: TileCacheService,
         feature_list_service: FeatureListService,
+        redis: Redis[Any],
     ):
         self.feature_table_cache_metadata_service = feature_table_cache_metadata_service
         self.namespace_handler = namespace_handler
@@ -67,6 +69,7 @@ class FeatureTableCacheService:
         self.entity_validation_service = entity_validation_service
         self.tile_cache_service = tile_cache_service
         self.feature_list_service = feature_list_service
+        self.redis = redis
 
     async def definition_hashes_for_nodes(
         self,
@@ -240,7 +243,7 @@ class FeatureTableCacheService:
         if is_target:
             return await get_target(
                 session=db_session,
-                redis=self.tile_cache_service.tile_manager_service.redis,
+                redis=self.redis,
                 graph=graph,
                 nodes=nodes_only,
                 observation_set=observation_table,
--- a/featurebyte/utils/async_helper.py
+++ b/featurebyte/utils/async_helper.py
@@ -246,7 +246,7 @@ def asyncio_gather(
             key=semaphore_key,
             limit=max_concurrency,
             timeout=LONG_RUNNING_EXECUTE_QUERY_TIMEOUT_SECONDS,
-        ) as semaphore:
+        ):
             nonlocal failed, tasks, tasks_canceled
             if failed:
                 # Close unawaited coroutines on failure
--- a/featurebyte/utils/async_helper.py
+++ b/featurebyte/utils/async_helper.py
@@ -120,7 +120,10 @@ class RedisFairCountingSemaphore:
 
 
 def asyncio_gather(
-    *coros: Coroutine[Any, Any, Any], redis: Redis[Any], max_concurrency: int = 0
+    *coros: Coroutine[Any, Any, Any],
+    redis: Redis[Any],
+    concurrency_key: str,
+    max_concurrency: int = 0,
 ) -> Future[List[Any]]:
     """
     Run coroutines with a optional concurrency limit
@@ -131,6 +134,8 @@ def asyncio_gather(
         Coroutines to run
     redis: Redis[Any]
         Redis connection
+    concurrency_key: str
+        Key for concurrency limit enforcement
     max_concurrency: int
         Maximum number of coroutines to run concurrently
 
@@ -145,11 +150,12 @@ def asyncio_gather(
     failed = False
     tasks_canceled = False
     tasks = []
+    semaphore_key = f"asyncio_gather:{concurrency_key}"
 
     # Run coroutines with a semaphore
     async def _coro(coro: Coroutine[Any, Any, Any]) -> Any:
         async with RedisFairCountingSemaphore(
-            redis=redis, key="asyncio_gather", limit=max_concurrency, timeout=3600
+            redis=redis, key=semaphore_key, limit=max_concurrency, timeout=3600
         ):
             nonlocal failed, tasks, tasks_canceled
             if failed:
--- a/featurebyte/utils/async_helper.py
+++ b/featurebyte/utils/async_helper.py
@@ -46,6 +46,7 @@ class RedisFairCountingSemaphore:
         self._ctr = f"{self._key}:counter"
         self._lock = f"{self._key}:lock"
         self._identifier: str | None = None
+        self._semaphore_refresh_task: Task[None] | None = None
 
     async def _acquire_semaphore(self) -> str | None:
         """
@@ -111,6 +112,10 @@ class RedisFairCountingSemaphore:
         if not self._identifier:
             return
 
+        # Cancel semaphore refresh task
+        if self._semaphore_refresh_task:
+            self._semaphore_refresh_task.cancel()
+
         # Release semaphore
         pipeline = self._redis.pipeline(True)
         pipeline.zrem(self._key, self._identifier)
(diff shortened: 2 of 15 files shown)


featurebyte/service/feature_table_cache.py
```diff
@@ -475,6 +476,7 @@ async def create_or_update_feature_table_cache(
                         column_names=[InternalName.TABLE_ROW_INDEX.value] + request_column_names,
                     ),
                     exists=True,
+                    retry=True,
```
YS-L, line 479: Is there a reason retry is needed here?

kchua78: without this change I often run into this error for Spark source type:

```
pyhive.exc.OperationalError: TExecuteStatementResp(status=TStatus(statusCode=3, infoMessages=['*org.apache.hive.service.cli.HiveSQLException:Error running query: io.delta.exceptions.ProtocolChangedException: The protocol version of the Delta table has been changed by a concurrent update. This happens when multiple writers are writing to an empty directory. Creating the table ahead of time will avoid this conflict. Please try the operation again.\
```

featurebyte/utils/async_helper.py
```diff
@@ -5,18 +5,137 @@
...
+        lock = self._redis.lock(self._lock, timeout=lock_timeout)
+        now = time.time()
+        while not lock.acquire(blocking=False):
+            if time.time() - now > self._timeout:
+                raise TimeoutError(f"Failed to acquire lock: {time.time() - now}")
+            await asyncio.sleep(0.1)
+
+        # Increment distributed counter
+        try:
+            counter = self._redis.incr(self._ctr)
```
YS-L, line 65: I read that the redis INCR command is atomic. Is the lock around it necessary?

featurebyte/utils/async_helper.py
```diff
@@ -31,11 +150,13 @@ def asyncio_gather(*coros: Coroutine[Any, Any, Any], max_concurrency: int = 0) -
...
     tasks_canceled = False
     tasks = []
-    semaphore = asyncio.Semaphore(max_concurrency)
+    semaphore_key = f"asyncio_gather:{concurrency_key}"
 
     # Run coroutines with a semaphore
     async def _coro(coro: Coroutine[Any, Any, Any]) -> Any:
-        async with semaphore:
+        async with RedisFairCountingSemaphore(
+            redis=redis, key=semaphore_key, limit=max_concurrency, timeout=3600
```
YS-L, line 158: Does this mean that a query will be queued for maximum of 1 hour by default before it raises an error? Is this a new behaviour compared to the existing behaviour?

featurebyte/utils/async_helper.py
```diff
@@ -5,18 +5,137 @@
...
+        pipeline = self._redis.pipeline(True)
+
+        # Remove expired semaphores
+        pipeline.zremrangebyscore(self._key, "-inf", now - self._timeout)
+        pipeline.zinterstore(self._czset, {self._czset: 1, self._key: 0})
+
+        # Acquire semaphore
+        pipeline.zadd(self._key, {self._identifier: now})
+        pipeline.zadd(self._czset, {self._identifier: counter})
+        pipeline.zrank(self._czset, self._identifier)
```
YS-L, line 82: Would be good to have some comments on the general idea of the implementation.

featurebyte/utils/async_helper.py
```diff
@@ -5,18 +5,137 @@
...
+        TimeoutError
+            If semaphore is not acquired within timeout
+        """
+        # Acquire lock for semaphore
+        lock_timeout = 1
+        lock = self._redis.lock(self._lock, timeout=lock_timeout)
+        now = time.time()
+        while not lock.acquire(blocking=False):
+            if time.time() - now > self._timeout:
+                raise TimeoutError(f"Failed to acquire lock: {time.time() - now}")
```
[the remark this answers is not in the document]
YS-L: Looking at this again I don't think this is possible and shouldn't be a concern.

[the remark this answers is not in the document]
kchua78: yes the call is non-blocking but atomic

featurebyte/utils/async_helper.py
kchua78: There is a mistake in the implementation. The lock is not meant to prevent race conditions in incrementing the counter, it's to avoid the situation that a process that got an earlier counter failing to acquire the semaphore due to another process that got a later counter acquiring it first (and hence violating the fairness of the semaphore).

The semaphore acquisition block should be done before releasing the lock.

featurebyte/utils/async_helper.py
gee-senbong: Would be good if you could share some reference links here too.

featurebyte/utils/async_helper.py
kchua78: In both the current and new behaviour there is a limit on how many SQL commands can be executed at the same time. 

Current behaviour: 
- the limit is local and more commands can be executed in parallel via multiple worker tasks
- no timeout on semaphore (since it's local) but there should be timeout from the calling function or the worker task

New behaviour: t
- the limit is cluster-wide and specific to each data warehouse
- timeout on semaphore to avoid situations where a process with the semaphore is terminated without releasing the semaphore indefinitely
- timeout from the calling function or the worker task still applies and will trigger release of the semaphore acquired

We can set the semaphore timeout to a value that is larger than expected higher level timeout limits to keep current behaviour

featurebyte/service/feature_table_cache.py
```diff
@@ -238,6 +238,7 @@ async def _populate_intermediate_table(
         if is_target:
             await get_target(
                 session=db_session,
+                redis=self.tile_cache_service.tile_manager_service.redis,
```
gee-senbong, line 241: nit: maybe it is cleaner to add the `redis` to this service & use it directly?

featurebyte/utils/async_helper.py
```diff
@@ -5,18 +5,137 @@
...
+        Acquire semaphore on enter, blocks until semaphore is acquired or timeout
+
+        Raises
+        ------
+        TimeoutError
+            If semaphore is not acquired within timeout
+        """
+        # Acquire lock for semaphore
+        lock_timeout = 1
+        lock = self._redis.lock(self._lock, timeout=lock_timeout)
```
gee-senbong, line 56: From the documentation:
```
        If specified, ``timeout`` indicates a maximum life for the lock.
        By default, it will remain locked until release() is called.
```
Seems like the operation here is about increment the counter, 1s should be sufficient in this case.

featurebyte/utils/async_helper.py
```diff
@@ -5,18 +5,137 @@
...
+        now = time.time()
+        while not lock.acquire(blocking=False):
+            if time.time() - now > self._timeout:
+                raise TimeoutError(f"Failed to acquire lock: {time.time() - now}")
+            await asyncio.sleep(0.1)
```
gee-senbong, line 61: Is the main purpose of not using `blocking_timeout` in `Redis.lock` is because of the block there is not async? Would be good add a documentation here so that we won't modify this part later & break something unintentionally.

featurebyte/utils/async_helper.py
kchua78: Looks like the lock should be released only after the semaphore is acquired, so it will need to be set to the same timeout as the semaphore

featurebyte/utils/async_helper.py
kchua78: Sure, will add a comment. Generally we should not be using non async blocking calls in async functions as that will block other threads from execution

featurebyte/utils/async_helper.py
gee-senbong: Ok. If i understood it correctly, we only need 1 timeout parameter in this case.

kchua78: it seems so. @YS-L Would it make sense to set it to `LONG_RUNNING_EXECUTE_QUERY_TIMEOUT_SECONDS`?

YS-L: I think that makes sense. This lock expiring doesn't mean that the query will get to run immediately, but just to get a "ticket" to wait for the semaphore, is that right?

kchua78: From what I understand this timeout should start when the lock is successfully acquired. When it times out the lock is released for other clients even if it's not released by the lock owner. 

If we set this as long as the lock owner execution lifespan (or longer) this should keep the lock mechanism intact. If this is too short then another process may acquire the lock before it's released. If it's set too long the it will block other processes from acquiring the lock for an extended period if the lock owner is terminated in a way that does not release the lock

featurebyte/utils/async_helper.py
```diff
@@ -5,18 +5,219 @@
...
+        str | None
+            Semaphore identifier if acquired, None otherwise
+        """
+        # Create a pipeline to execute multiple commands atomically
+        pipeline = self._redis.pipeline(True)
+
+        # Remove expired semaphore records from the identity and owner zsets
+        now = time.time()
+        pipeline.zremrangebyscore(self._key, "-inf", now - SEMAPHORE_TIMEOUT_SECONDS)
+        pipeline.zinterstore(self._czset, {self._czset: 1, self._key: 0})
```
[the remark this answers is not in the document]
gee-senbong: Oh.. I understand it based on the GPT explanation:
```
Intersecting Sets: zinterstore updates self._czset to keep only entries present in both self._czset and self._key. This removes identifiers that have expired.
```

After removing the expired items from the `identity zset`, it takes the intersection between `owner zset` & `identity zset` and store the results in the `owner set`.

featurebyte/utils/async_helper.py
```diff
@@ -5,18 +5,219 @@
...
+
+        # Get queue position from redis counter
+        pipeline.incr(self._ctr)
+        counter = pipeline.execute()[-1]
```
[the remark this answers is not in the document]
kchua78: Only the new statements get executed in the next execute call

featurebyte/utils/async_helper.py
```diff
@@ -5,18 +5,219 @@
...
+
+    Parameters
+    ----------
+    task: Task[Any]
+        Task to monitor
+    semaphore: RedisFairCountingSemaphore
+        Semaphore to refresh
+    """
+    # Refresh semaphore while task is running
+    while not task.done() and not task.cancelled():
```
gee-senbong, line 199: If the task is an IO task, just want to check whether `task.done()` & `task.cancelled()` works properly in gevent environment.

gee-senbong (approved):
LGTM

featurebyte/utils/async_helper.py
kchua78: It should. Gevent patches blocking system calls which should not be used by asyncio.
I am also going to refactor this so the refresh task in handled in side the semaphore class, which should no longer require these checks as well

featurebyte/utils/async_helper.py
```diff
@@ -249,8 +254,6 @@ async def _coro(coro: Coroutine[Any, Any, Any]) -> Any:
                 return
             task = asyncio.create_task(coro)
             tasks.append(task)
-            # Add task to refresh semaphore while task is running
-            semaphore_refresh_task = asyncio.create_task(_semaphore_refresh(task, semaphore))
```
kchua78: Moved handling of the refresh task into the semaphore class so it's self contained

YS-L (approved):

PR #2615, opened 2024-10-01
[DEV-3745] Cancel sql query on timeout or task cancellation

kchua78:
## Description

Cancel sql query on timeout or task cancellation to avoid unnecessary data warehouse costs.

## Type of Change

## Checklist

### DIFF
--- a/featurebyte/query_graph/node/schema.py
+++ b/featurebyte/query_graph/node/schema.py
@@ -338,6 +338,10 @@ class BigQueryDetails(BaseDatabaseDetails):  # pylint: disable=abstract-method
     dataset_name: StrictStr = Field(
         description="The name of the dataset to use for creation of output tables."
     )
+    location: Optional[StrictStr] = Field(
+        default="US",
+        description="The location of the dataset to use for creation of output tables.",
+    )
 
     def get_source_info(self) -> SourceInfo:
         return SourceInfo(
--- a/featurebyte/session/base.py
+++ b/featurebyte/session/base.py
@@ -123,7 +123,7 @@ async def to_thread(
     except (asyncio.exceptions.TimeoutError,) + TaskRevokeExceptions:
         if error_handler:
             try:
-                await error_handler(*args, **kwargs)
+                await error_handler()
             except Exception:
                 logger.error("Error handling exception", exc_info=True)
 
@@ -557,7 +557,7 @@ class BaseSession(BaseModel):
             await to_thread(
                 self._execute_query,
                 timeout,
-                lambda *args, **kwargs: self._try_cancel_query(cursor, query),
+                lambda: self._try_cancel_query(cursor, query),
                 cursor,
                 query,
                 **self.get_additional_execute_query_kwargs(),
--- a/featurebyte/session/base.py
+++ b/featurebyte/session/base.py
@@ -29,7 +29,11 @@ from sqlglot.expressions import Expression, Select
 from featurebyte.common.path_util import get_package_root
 from featurebyte.common.utils import create_new_arrow_stream_writer, dataframe_from_arrow_stream
 from featurebyte.enum import InternalName, MaterializedTableNamePrefix, SourceType, StrEnum
-from featurebyte.exception import DataWarehouseOperationError, QueryExecutionTimeOut
+from featurebyte.exception import (
+    DataWarehouseOperationError,
+    QueryExecutionTimeOut,
+    TaskRevokeExceptions,
+)
 from featurebyte.logging import get_logger
 from featurebyte.models.user_defined_function import UserDefinedFunctionModel
 from featurebyte.query_graph.model.column_info import ColumnSpecWithDescription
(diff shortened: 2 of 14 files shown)


featurebyte/query_graph/node/schema.py
```diff
@@ -338,6 +338,10 @@ class BigQueryDetails(BaseDatabaseDetails):  # pylint: disable=abstract-method
     dataset_name: StrictStr = Field(
         description="The name of the dataset to use for creation of output tables."
     )
+    location: Optional[StrictStr] = Field(
```
kchua78, line 341: Location needs to specified for query cancellation

featurebyte/session/base.py
```diff
@@ -87,25 +95,41 @@ def _func_wrapper(func: Any, thread_info: Dict[str, Any], *args: Any, **kwargs:
...
-            logger.debug("Raised exception in thread")
+            logger.info("Raised exception in thread")
 
     thread_info: Dict[str, int] = {}
     func_call = functools.partial(ctx.run, _func_wrapper, func, thread_info, *args, **kwargs)
 
     try:
-        return await asyncio.wait_for(loop.run_in_executor(None, func_call), timeout)
-    except asyncio.exceptions.TimeoutError:
+        coro = loop.run_in_executor(None, func_call)
```
kchua78, line 112: Check for task completion without using await or wait_for so that task cancellation can be caught

featurebyte/session/bigquery.py
```diff
@@ -284,6 +287,25 @@ def get_additional_execute_query_kwargs(self) -> dict[str, Any]:
         )
         return {"job_config": job_config}
 
+    @property
+    def dataset_ref(self) -> DatasetReference:
+        return DatasetReference(project=self.project_name, dataset_id=self.dataset_name)
+
+    @staticmethod
+    def _execute_query(cursor: Any, query: str, **kwargs: Any) -> Any:
+        job_id = str(uuid.uuid4())
```
kchua78, line 296: specify job id for request so we don't need to wait for job handle to be available after query

featurebyte/session/bigquery.py
```diff
@@ -364,12 +386,10 @@ def _get_schema_from_cursor(self, cursor: Any) -> pa.Schema:
         -------
         pa.Schema
         """
-        # get schema from query job as cursor description is missing REPEATED information
-        try:
-            schema = self._client._get_query_results(
-                cursor.query_job.job_id, DEFAULT_RETRY, page_size=1
-            ).schema  # pylint: disable=protected-access
-        except NotFound:
+        rows = cursor._query_rows  # pylint: disable=protected-access
```
kchua78, line 389: This will avoid unnecessary api call to get schema

featurebyte/session/bigquery.py
```diff
@@ -597,30 +617,30 @@ async def get_table_details(
         database_name: str | None = None,
         schema_name: str | None = None,
     ) -> TableDetails:
-        query = (
```
kchua78, line 600: Using `INFORMATION_SCHEMA` is not reliable as it will fail when the dataset location does not match the client location

featurebyte/session/databricks.py
```diff
@@ -50,15 +71,22 @@ def _initialize_connection(self) -> None:
...
-        self._connection = databricks_sql.connect(
-            server_hostname=self.host,
-            http_path=self.http_path,
-            access_token=self.database_credential.access_token,
-            catalog=self.catalog_name,
-            schema=self.schema_name,
-            _user_agent_entry=APPLICATION_NAME,
-            _use_arrow_native_complex_types=False,
-        )
+        with patch("databricks.sql.client.ThriftBackend", ThriftBackend):
```
kchua78, line 74: This patch is necessary for the query execution to be cancellable while the cursor is polling for results

featurebyte/session/base.py
```diff
@@ -490,7 +559,12 @@ async def get_async_query_stream(
         try:
             # execute in separate thread
             await to_thread(
-                cursor.execute, timeout, query, **self.get_additional_execute_query_kwargs()
+                self._execute_query,
+                timeout,
+                lambda cursor, query, **kwargs: self._try_cancel_query(cursor, query),
```
YS-L, line 564: Does this lambda need to have `cursor` and `query` as the arguments or can it simply be:

```
lambda: self._try_cancel_query(cursor, query)
```

featurebyte/session/bigquery.py
YS-L: Would be good to include this comment in the code.

featurebyte/session/databricks.py
YS-L: Would be good to include this as a comment in the code.

featurebyte/session/base.py
```diff
@@ -238,6 +262,47 @@ def connection(self) -> Any:
...
+        Returns
+        -------
+        bool
+            True if query was cancelled, False otherwise
+        """
+        try:
+            cursor.cancel()
+            return True
+        except AttributeError:
+            return False
```
YS-L, line 285: It seems almost all sessions have to override this implementation. In which session type would this default implementation work?

featurebyte/session/bigquery.py
YS-L: Is there any concern over this note on setting the job id manually? https://github.com/googleapis/python-bigquery/blob/e29b987956e5d80541ad9a573e902938a1373bda/google/cloud/bigquery/dbapi/cursor.py#L180-L183

featurebyte/session/base.py
kchua78: Indeed. That should work

featurebyte/session/bigquery.py
kchua78: When not provided the bigquery client code will create one the same way as here:
https://github.com/googleapis/python-bigquery/blob/e29b987956e5d80541ad9a573e902938a1373bda/google/cloud/bigquery/_job_helpers.py#L84

featurebyte/session/base.py
kchua78: There is SQLiteSession. Would make sense to make this an abstractmethod and add an implementation there?

YS-L: I think would be good to make this an abstractmethod since most likely it needs to have a custom implementation.

featurebyte/session/base.py
```diff
@@ -490,7 +555,12 @@ async def get_async_query_stream(
         try:
             # execute in separate thread
             await to_thread(
-                cursor.execute, timeout, query, **self.get_additional_execute_query_kwargs()
+                self._execute_query,
+                timeout,
+                lambda *args, **kwargs: self._try_cancel_query(cursor, query),
```
YS-L, line 560: Are `args` and `kwargs` needed here? If not, might be better to make the lambda have no argument and call the handler without passing anything in.

YS-L (approved):

featurebyte/session/base.py
kchua78: probably not. Will remove for now - we can update this later if we need the callback to use any other information not available before code is run (e.g. exception)

PR #2626, opened 2024-10-10
[DEV-3801] Extend feature table cache to multiple tables

YS-L:
## Description

This extends feature table cache to multiple tables to avoid issues when tables become too wide.

By default, each feature cache table will store up to 1000 columns. New feature cache tables will be created if the limit is exceeded. The retrieval of features from feature table cache will join the tables by row index if necessary.

## Type of Change

## Checklist


featurebyte/service/feature_table_cache_metadata.py
```diff
@@ -57,64 +63,113 @@ def __init__(
...
-            documents.append(document)
 
-        if documents:
-            document = documents[0]
-        else:
+        eligible_cache_metadata = None
+        num_cache_tables = 0
+        async for cache_metadata in self.list_documents_iterator(query_filter=query_filter):
+            if (
+                len(cache_metadata.feature_definitions) + num_columns_to_insert
```
kchua78, line 122: `num_columns_to_insert` should be much smaller than `FEATUREBYTE_FEATURE_TABLE_CACHE_MAX_COLUMNS` right?

featurebyte/service/feature_table_cache_metadata.py
```diff
@@ -57,64 +63,113 @@ def __init__(
...
+            ) <= FEATUREBYTE_FEATURE_TABLE_CACHE_MAX_COLUMNS:
+                eligible_cache_metadata = cache_metadata
+            num_cache_tables += 1
+
+        if eligible_cache_metadata is None:
```
kchua78, line 127: For existing observation tables with a feature cache table we will assume there is no cache table?

featurebyte/service/feature_table_cache_metadata.py
YS-L: Currently `num_columns_to_insert` depends on the number of uncached features in a single request.

Usually it should be smaller than `FEATUREBYTE_FEATURE_TABLE_CACHE_MAX_COLUMNS`, but when it is larger the current implementation won't split the table but instead will create a new table that has more columns than `FEATUREBYTE_FEATURE_TABLE_CACHE_MAX_COLUMNS`. This shouldn't break anything unless `num_columns_to_insert` is significantly larger than the limit e.g. by 10x.

We can perform batching to ensure `num_columns_to_insert` is not too large as an improvement.

featurebyte/service/feature_table_cache_metadata.py
YS-L: Yes, the existing cache is still valid.

featurebyte/service/feature_table_cache_metadata.py
kchua78: I guess it should be fine since it will be a larger table than expected but does not need to be merged into, which is what we are trying to avoid here

kchua78 (approved):

PR #2633, opened 2024-10-14
[DEV-3810] Fix entity untagging bug when multiple paths exist

gee-senbong:
## Description

Fix the issue where entity untagging removed the path that can be reached by other parents.

## Type of Change

## Checklist

### DIFF
--- a/featurebyte/service/relationship.py
+++ b/featurebyte/service/relationship.py
@@ -5,7 +5,7 @@ RelationshipService class
 from __future__ import annotations
 
 from abc import abstractmethod
-from typing import TypeVar, cast
+from typing import Optional, Set, TypeVar, cast
 
 from bson import ObjectId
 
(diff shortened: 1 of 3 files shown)


tests/unit/service/test_relationship.py
```diff
@@ -428,3 +434,109 @@ async def test_remove_relationship__case_3(
...
+    )
+    await family_relationship_service.add_relationship(
+        parent=Parent(id=object_d.id), child_id=object_e.id
+    )
+
+    expected_msg = 'Object "object_a" is already an ancestor of object "object_d".'
+    with pytest.raises(DocumentUpdateError, match=expected_msg):
+        await family_relationship_service.add_relationship(
+            parent=Parent(id=object_a.id), child_id=object_d.id
+        )
```
gee-senbong, line 524: Added this test so that we are aware of the implication if we support this case in the future.

featurebyte/service/relationship.py
```diff
@@ -138,15 +138,65 @@ async def add_relationship(self, parent: ParentT, child_id: ObjectId) -> Relatio
...
+                ancestor_ids=ancestor_ids,
+                parents=keep_parents,
+            ),
+        )
+        checked_descendant_ids.add(document.id)
+
+        async for descendant_obj in self.document_service.list_documents_iterator(
+            query_filter={"parents.id": document.id},
+        ):
+            await self._remove_ancestor_ids(
```
YS-L, line 192: Curious what is the expected number of queries this function will call since it's recursive?

YS-L (approved):

featurebyte/service/relationship.py
```diff
@@ -155,52 +205,41 @@ async def remove_relationship(self, parent_id: ObjectId, child_id: ObjectId) ->
...
-        assert isinstance(child_object, Relationship)
-        self._validate_remove_relationship_operation(
-            parent_obj=parent_object, child_obj=child_object
-        )
+        parent_obj = await self.document_service.get_document(document_id=parent_id)
+        child_obj = await self.document_service.get_document(document_id=child_id)
+        assert isinstance(parent_obj, Relationship), "Parent object is not a Relationship object"
+        assert isinstance(child_obj, Relationship), "Child object is not a Relationship object"
 
-        async with self.persistent.start_transaction():
```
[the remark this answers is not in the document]
gee-senbong: Yes, if not it likely causes the transaction timeout issue.

featurebyte/service/relationship.py
gee-senbong: We don't expect the number to be high. It is scaled with number of descendent of a particular entity where its parent get removed.

gee-senbong: Before current changes, there are like 2 update queries
* 1 update current parent-child entity relationship
* 1 update descendent of the current child entity (to update descendants' ancestor IDs)

After current changes, the number of queries is scaled with number of descendants that the parent entity has.

kchua78: If we change the filter condition to `{"ancestor_ids": document.id}` then this function does not need to be recursive?

gee-senbong: Do you mean making multiple updates inside the non-recursive function?

kchua78: Looks like we need to do this recursively since there is no restriction on shared ancestry from different parents

kchua78 (approved):

PR #2635, opened 2024-10-16
Revert "[chore] Standardized function naming"

bushwhackr:
Reverts featurebyte/featurebyte#2627

Adding group_ids to credentials


tests/unit/conftest.py
```diff
@@ -725,6 +725,7 @@ def credentials_fixture(snowflake_feature_store_params):
     return {
         snowflake_feature_store_params["name"]: CredentialModel(
             name="sf_featurestore",
+            group_id=[],
```
kchua78, line 728: Should this be `group_ids`

PR #2692, opened 2024-12-05
[DEV-3913] Fix observation table creation failure when datetime values or sampling contain timezone info

kchua78:
## Description

This PR makes the following changes during observation table materialization:
- Cast `sample_from_timestamp` and `sample_to_timestamp` to UTC timestamps without timezone info
- Exclude rows where mapped columns contain NULL values
- Exclude rows where `POINT_IN_TIME` is too recent for feature materialization

## Type of Change

## Checklist

### DIFF
--- a/featurebyte/models/request_input.py
+++ b/featurebyte/models/request_input.py
@@ -5,14 +5,15 @@ RequestInput is the base class for all request input types.
 from __future__ import annotations
 
 from abc import abstractmethod
-from datetime import datetime
+from datetime import datetime, timedelta
 from typing import Any, Dict, List, Literal, Optional, cast
 
+from dateutil import tz
 from pydantic import Field, PrivateAttr, StrictStr
 from sqlglot import expressions
 from sqlglot.expressions import Select
 
-from featurebyte.enum import SpecialColumnName, StrEnum
+from featurebyte.enum import DBVarType, SpecialColumnName, StrEnum
 from featurebyte.exception import ColumnNotFoundError
 from featurebyte.models.base import FeatureByteBaseModel
 from featurebyte.query_graph.model.common_table import TabularSource
@@ -30,6 +31,8 @@ from featurebyte.query_graph.sql.materialisation import (
 from featurebyte.query_graph.sql.source_info import SourceInfo
 from featurebyte.query_graph.transform.operation_structure import OperationStructureExtractor
 from featurebyte.session.base import BaseSession
+
+HISTORICAL_REQUESTS_POINT_IN_TIME_RECENCY_HOUR = 48
 
 
 class RequestInputType(StrEnum):
@@ -88,9 +91,9 @@ class BaseRequestInput(FeatureByteBaseModel):
         return int(result.iloc[0]["row_count"])  # type: ignore[union-attr]
 
     @abstractmethod
-    async def get_column_names(self, session: BaseSession) -> list[str]:
+    async def get_column_names_and_dtypes(self, session: BaseSession) -> Dict[str, DBVarType]:
         """
-        Get the column names of the table query
+        Get the column names and dtypes of the table query
 
         Parameters
         ----------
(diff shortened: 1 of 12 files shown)


featurebyte/models/request_input.py
```diff
@@ -148,20 +151,50 @@ async def materialize(
...
+            columns = available_columns
+        else:
+            columns = self.columns
+
+        if self.columns_rename_mapping is not None:
+            output_columns = [self.columns_rename_mapping.get(col, col) for col in columns]
+            column_names_and_dtypes = {
+                self.columns_rename_mapping.get(col, col): dtype
+                for col, dtype in column_names_and_dtypes.items()
+            }
```
gee-senbong, line 167: nit: Do we need to consider the possibility of column name conflict here (user renames the column to existing column name which causes more than 1 keys with the same name).
            }

gee-senbong (approved):

featurebyte/models/request_input.py
YS-L: This is an existing issue. I think we could update https://github.com/featurebyte/featurebyte/blob/23f3921586941cd3234dc08e698463b91ddad979/featurebyte/models/request_input.py#L214-L223 to prevent column name conflict.

featurebyte/models/request_input.py
```diff
@@ -182,7 +219,22 @@ async def materialize(
...
+                        this=quoted_identifier(col_name),
+                        expression=expressions.Not(this=expressions.Null()),
+                    )
+                )
+        if filter_conditions:
+            # if the expression is a select * from table, we need to select the columns explicitly
+            if query_expr.expressions and query_expr.expressions[0].name == "*":
+                query_expr = select_and_rename_columns(
+                    query_expr, columns, self.columns_rename_mapping
+                )
```
YS-L, line 236: Is it possible that `select_and_rename_columns` is already called once above? If so will there be issue if the columns rename mapping is applied again?

featurebyte/models/request_input.py
```diff
@@ -148,20 +151,50 @@ async def materialize(
...
+
+        if SpecialColumnName.POINT_IN_TIME in output_columns and column_names_and_dtypes[
+            SpecialColumnName.POINT_IN_TIME
+        ] in {
+            DBVarType.TIMESTAMP,
+            DBVarType.TIMESTAMP_TZ,
+        }:
+            max_timestamp = datetime.utcnow() - timedelta(
+                hours=HISTORICAL_REQUESTS_POINT_IN_TIME_RECENCY_HOUR
+            )
```
YS-L, line 179: `RequestInput.materialize` is called not just when creating observation table, but also when creating batch request table and source static source table:

image

The new logic here is only applicable when creating observation table.

YS-L: I think a cleaner approach would be to achieve this behaviour through parameters passed to `materialize` in observation table creation task:

* `sample_to_timestamp`: update this value to account for `HISTORICAL_REQUESTS_POINT_IN_TIME_RECENCY_HOUR` before calling the method
* `columns_exclude_missing_values`: List of columns that should not have missing values (we know what columns will be entity columns)

featurebyte/models/request_input.py
```diff
@@ -148,93 +147,85 @@ async def materialize(
...
-        else:
-            max_timestamp = None
+            output_column_names_and_dtypes = column_names_and_dtypes
+            output_columns = input_columns
 
-        if self.columns is not None or self.columns_rename_mapping is not None:
-            self._validate_columns_and_rename_mapping(available_columns)
-            query_expr = select_and_rename_columns(query_expr, columns, self.columns_rename_mapping)
+        # always perform explicit column selection and renaming
+        query_expr = select_and_rename_columns(
```
kchua78, line 171: This will ensure the query expression include explicit column selection

YS-L (approved):

PR #2709, opened 2024-12-13
[DEV-3910] Update catalog cleanup to delete audit logs

gee-senbong:
## Description

This PR revises catalog cleanup task so that
* it deletes audit log documents
* it revises cleanup task structure so that it can be extended to handle assets without `catalog_id`

## Type of Change

## Checklist

### DIFF
--- a/featurebyte/worker/task/catalog_cleanup.py
+++ b/featurebyte/worker/task/catalog_cleanup.py
@@ -173,6 +173,7 @@ class CatalogCleanupTask(BaseTask[CatalogCleanupTaskPayload]):
     async def _pre_cleanup_non_catalog_mongo_asset(
         self,
         catalog: CatalogModel,
+        progress_callback: ProgressCallbackType,
     ) -> None:
         """
         Handle the cleanup of non-catalog mongo assets before the catalog specific assets are cleaned up.
@@ -183,17 +184,18 @@ class CatalogCleanupTask(BaseTask[CatalogCleanupTaskPayload]):
         ----------
         catalog: CatalogModel
             Catalog model object
+        progress_callback: ProgressCallbackType
+            Progress callback
         """
 
     async def _post_cleanup_non_catalog_mongo_asset(
         self,
         catalog: CatalogModel,
         progress_callback: ProgressCallbackType,
-        catalog_collection_percent: int,
     ) -> None:
         # cleanup user defined functions
         await progress_callback(
-            percent=catalog_collection_percent + 1,
+            percent=30,
             message="Cleaning up user defined functions",
         )
 
@@ -205,7 +207,7 @@ class CatalogCleanupTask(BaseTask[CatalogCleanupTaskPayload]):
 
         # cleanup task documents
         await progress_callback(
-            percent=catalog_collection_percent + 2,
+            percent=60,
             message="Cleaning up task documents",
         )
         await self.persistent.delete_many(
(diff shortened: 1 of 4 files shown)


featurebyte/worker/task/catalog_cleanup.py
```diff
@@ -214,44 +286,24 @@ async def _cleanup_catalog(
                 percent=int(100 * i / len(self.catalog_specific_model_class_pairs)),
                 message="Cleaning up Mongo records",
             )
-            await self.persistent.delete_many(
+            with_audit = True
+            model_settings = getattr(model_class, "Settings", None)
+            if model_settings:
+                with_audit = getattr(model_settings, "auditable", True)
```
YS-L, line 292: Is there a reason to use `getattr` instead of directly accessing the attributes? If the auditable attribute doesn't exist, is it intended to assume it is auditable?

featurebyte/worker/task/catalog_cleanup.py
```diff
@@ -169,16 +170,87 @@ async def _cleanup_store_files(self, remote_file_paths: set[Path]) -> None:
...
 
+    async def _pre_cleanup_non_catalog_mongo_asset(
+        self,
+        catalog: CatalogModel,
+        progress_callback: ProgressCallbackType,
+    ) -> None:
+        """
+        Handle the cleanup of non-catalog mongo assets before the catalog specific assets are cleaned up.
+        This is also used by those assets that is catalog specific but it does not have a corresponding
+        catalog ID in the document.
```
YS-L, line 181: Curious what are some examples of such assets?

featurebyte/worker/task/catalog_cleanup.py
gee-senbong: For certain class like table model, we don't pass class object. We only pass the function used to construct the object. In this case, the object does not have `Settings` class.

featurebyte/worker/task/catalog_cleanup.py
gee-senbong: Feature list builder does not have catalog ID. To handle that case, we need to retrieve entity IDs of the catalog & then retrieve all the feature list builder records with match entity IDs.

YS-L: Ideally, feature list builder should be storing a catalog ID, is that right?

YS-L (approved):

tests/unit/worker/task/test_catalog_cleanup.py
```diff
@@ -144,6 +162,11 @@ async def test_catalog_cleanup(mock_drop_table, catalog_with_associated_document
     catalog_cleanup_task = app_container.catalog_cleanup_task
     catalog_service = app_container.catalog_service
     await catalog_service.soft_delete_document(document_id=catalog.id)
+
+    # add a delay to ensure the soft delete is completed
+    await asyncio.sleep(1)
```
YS-L, line 167: Is this sleep necessary? My understanding is that the `soft_delete_document` above is done synchronously.

featurebyte/worker/task/catalog_cleanup.py
gee-senbong: Yes, ideally it should store catalog ID. 😅

tests/unit/worker/task/test_catalog_cleanup.py
gee-senbong: When i run this locally (for this test), it is quite flaky. If I run the whole unit tests, it passes without any issue. After adding this delay, the issue gone.

featurebyte/worker/task/catalog_cleanup.py
YS-L: I see. I wonder if it's better to perform a migration to add the catalog ID instead of maintaining special logic like this.

gee-senbong: 🆗  We can add the migration script for the feature list builder at featurebyte-app to handle this.

PR #2711, opened 2024-12-14
[DEV-3922] Support non-timestamp special columns in SCDTable

YS-L:
## Description

This adds support for using non-timestamp columns as special columns in SCDTable.

## Type of Change

## Checklist


featurebyte/common/validator.py
```diff
@@ -68,6 +82,33 @@ def _model_validator(self: Any) -> Any:
...
+                timestamp_schema_mapping[col_name] = timestamp_schema
+            col_dtype = col_info_map[col_name].get("dtype")
+            if col_dtype in ambiguous_timestamp_types and not timestamp_schema:
+                raise ValueError(
+                    f"timestamp_schema is required for {col_name} with ambiguous timestamp type {col_dtype}"
+                )
+
+        # Update columns_info with timestamp_schema
+        if timestamp_schema_mapping:
+            for col_info in col_info_map.values():
```
kchua78, line 106: :nit it seems that `col_info["name"]` is the same as the keys of `col_info_map`. Would it be more efficient to iterate `timestamp_schema_mapping` instead?

featurebyte/query_graph/node/schema.py
```diff
@@ -412,3 +412,12 @@ class ColumnSpec(FeatureByteBaseModel):
     name: NameStr
     dtype: DBVarType
     timestamp_schema: Optional[TimestampSchema] = None
+
+    @model_validator(mode="after")
+    def _validate_string_timestamp_format_string(self) -> "ColumnSpec":
+        if self.dtype == DBVarType.VARCHAR and self.timestamp_schema is not None:
```
kchua78, line 418: should we also raise an error if `self.timestamp_schema` is None when the dtype is VARCHAR?

featurebyte/query_graph/sql/ast/input.py
```diff
@@ -162,6 +167,63 @@ def make_input_columns_map(cls, context: SQLNodeContext) -> dict[str, Expression
...
+        column_expr: Expression = expressions.Identifier(this=column_spec.name, quoted=True)
+
+        if column_spec.dtype == DBVarType.VARCHAR:
+            assert timestamp_schema.format_string is not None
+            column_expr = adapter.to_timestamp_from_string(
+                column_expr, timestamp_schema.format_string
+            )
+        elif column_spec.dtype == DBVarType.DATE:
+            # Treat date type columns as end of day in the local timezone
+            column_expr = adapter.dateadd_second(make_literal_value(86400), column_expr)
```
kchua78, line 218: Should this be `86399`? If we use strict less than for cutoffs then this should be ok

featurebyte/query_graph/node/schema.py
YS-L: Regular string columns will have `VARCHAR` dtype and `timestamp_schema` of None. We should raise this error only when the string column is a special datetime column, so it cannot be validated at the `ColumnSpec` level but at the table level.

This is currently being validated here in this PR: https://github.com/featurebyte/featurebyte/pull/2711/files#diff-dd1960347f0045df362d18aee725acbe91a1f6c6b01af116d58bfa91dcb5494fR99-R102

featurebyte/query_graph/sql/ast/input.py
YS-L: For SCD joins we use `<=` condition for cutoffs. 

For a row in an SCD table with an effective date value of `2024-01-15`, what is the earliest point in time that can be used to lookup that row?

1. `2024-01-16 00:00:00`
2. `2024-01-15 23:59:59`

The current version implements 1). Using `86399` would be 2). I thought 1) is cleaner and more conservative. Either way, only the point-in-times within that one second are affected. Let me know if you have a preference and if I should change it.

kchua78: I think we can stick with option 1

kchua78 (approved):

PR #2849, opened 2025-02-26
[chore] Fix batch feature table recreate fails when parent table is deleted

kchua78:
## Description

Batch feature table recreate tries to get name from parent table which can fail if the parent table is deleted.

This PR make the following changes:
- Store the name of the parent batch feature table instead of the identifier so the original parent table does not need to exist to generate the name of the new batch feature table
- Update the uniqueness constraints of batch feature table name so that the name only needs to be unique for a specific deployment

## Type of Change

## Checklist

### DIFF
--- a/featurebyte/models/batch_feature_table.py
+++ b/featurebyte/models/batch_feature_table.py
@@ -4,12 +4,16 @@ BatchFeatureTable related model(s)
 
 from __future__ import annotations
 
-from typing import Optional
+from typing import List, Optional
 
 import pymongo
 from pydantic import Field
 
-from featurebyte.models.base import PydanticObjectId
+from featurebyte.models.base import (
+    PydanticObjectId,
+    UniqueConstraintResolutionSignature,
+    UniqueValuesConstraint,
+)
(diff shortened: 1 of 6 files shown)


featurebyte/models/batch_feature_table.py
```diff
@@ -22,7 +26,7 @@ class BatchFeatureTableModel(MaterializedTableModel):
     batch_request_table_id: Optional[PydanticObjectId]
     request_input: Optional[BatchRequestInput] = Field(default=None)
     deployment_id: PydanticObjectId
-    parent_batch_feature_table_id: Optional[PydanticObjectId] = Field(default=None)
```
YS-L, line 25: Will it still be useful to track the parent feature table id for other purposes?

YS-L (approved):

featurebyte/models/batch_feature_table.py
kchua78: Currently there is no use for that information but we can re-introduce it if we need it in the future

PR #2860, opened 2025-03-05
[DEV-4002] Disable online feature serving if online store is not configured for the catalog

kchua78:
## Description

Disable online feature serving if online store is not configured for the catalog

## Type of Change

## Checklist


tests/unit/routes/test_deployment.py
```diff
@@ -41,6 +41,17 @@ def _mock_online_enable_service_update_data_warehouse(self, mock_deployment_flow
...
     @pytest.fixture(autouse=True)
     def _always_patch_app_get_storage_fixture(self, storage):
+        """
+        Patch app.get_storage for all tests in this module
+        """
+        with patch(
+            "featurebyte.routes.deployment.controller.DeploymentController._validate_deployment_is_online_enabled",
+            return_value=None,
+        ):
+            yield
```
YS-L, line 51: nit: I think this is supposed to be under `_always_patch_deployment_online_fixture`?

featurebyte/routes/deployment/controller.py
```diff
@@ -277,13 +282,33 @@ async def compute_online_features(
...
+        Raises
+        ------
+        DeploymentNotOnlineEnabledError
+            Deployment is not online enabled
+        """
+        catalog = await self.catalog_service.get_document(document_id=deployment.catalog_id)
+        if catalog.online_store_id is None:
+            raise DeploymentNotOnlineEnabledError(
+                "Deployment is not online enabled. Configure online store for the catalog to enable online feature requests."
+            )
```
YS-L, line 310: Is this scenario covered by a test?

featurebyte/routes/deployment/controller.py
```diff
@@ -277,13 +282,33 @@ async def compute_online_features(
...
+        deployment: DeploymentModel
+            Deployment to validate
+
+        Raises
+        ------
+        DeploymentNotOnlineEnabledError
+            Deployment is not online enabled
+        """
+        catalog = await self.catalog_service.get_document(document_id=deployment.catalog_id)
+        if catalog.online_store_id is None:
```
YS-L, line 307: Do we have to handle the case where the online store is unset in the catalog after deployment is enabled?

kchua78: In that case the deployment is still enabled and supports batch feature tables but not online serving. Do you anticipate some unexpected behaviour?

featurebyte/routes/deployment/controller.py
kchua78: Will add one

featurebyte/routes/deployment/controller.py
YS-L: In that case, the deployment would still be supported by scheduled feature jobs. So while the deployment is still capable of low latency serving, the logic here would not allow that. Or is the expectation that when catalog is unset, we should stop the feature jobs from updating the online store?

YS-L: After double checking I confirm the online stores are no longer updated when catalog's online store is unset, so this  should the the expected behaviour.

PR #2901, opened 2025-04-01
[chore] Revise table sampling logic to cache sampled primary table

gee-senbong:
## Description

This PR revises table sampling logic to cached sampled primary table to improve runtime performance.

## Type of Change

## Checklist

### DIFF
--- a/featurebyte/query_graph/sql/interpreter/preview.py
+++ b/featurebyte/query_graph/sql/interpreter/preview.py
@@ -203,45 +203,6 @@ class PreviewMixin(BaseGraphInterpreter):
             ),
         )
 
-    def _get_query_graph_with_sample_on_primary_input_nodes(
-        self,
-        sample_on_primary_table: bool,
-        target_node_name: str,
-        seed: int,
-        sort_by_prob: bool,
-        sample_row_num: Optional[int] = None,
-        total_num_rows: Optional[int] = None,
-    ) -> tuple[QueryGraph, bool]:
-        """
-        Get query graph with sample row number set on primary input nodes
-
-        Parameters
-        ----------
-        sample_on_primary_table: bool
-            Whether to sample on primary table
-        target_node_name: str
-            Target node name
-        seed: int
-            Random seed
-        sort_by_prob: bool
-            Whether to sort sampled result by the random probability to correct for oversampling
-            bias. Can be expensive on large samples.
-        sample_row_num: Optional[int]
-            Number of rows to sample
-        total_num_rows: Optional[int]
-            Total number of rows before sampling
-
-        Returns
-        -------
-        Tuple[QueryGraph, bool]
-            Query graph with sample row number set on primary input nodes, whether input node is sampled
-        """
-        query_graph = QueryGraph(**self.query_graph.model_dump(by_alias=True))
-        input_node_sampled = False
-        if not sample_on_primary_table:
-            return query_graph, input_node_sampled
-        return query_graph, True
-
     def _construct_sample_sql(  # pylint: disable=too-many-arguments,too-many-locals
         self,
         node_name: str,
--- a/featurebyte/query_graph/sql/interpreter/preview.py
+++ b/featurebyte/query_graph/sql/interpreter/preview.py
@@ -243,7 +243,7 @@ class PreviewMixin(BaseGraphInterpreter):
         clip_timestamp_columns: bool
             Whether to apply clipping to all the timestamp columns
         sample_on_primary_table: bool
-            Whether to sample on primary table
+            Whether the graph is already sampled on primary table
 
         Returns
         -------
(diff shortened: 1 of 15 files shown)


featurebyte/service/preview.py
```diff
@@ -152,6 +154,50 @@ async def _get_or_cache_dataframe(
...
+            num_rows=num_rows,
+            seed=seed,
+            from_timestamp=sample.from_timestamp,
+            to_timestamp=sample.to_timestamp,
+            timestamp_column=sample.timestamp_column,
+            total_num_rows=total_num_rows,
+            skip_conversion=True,  # skip conversion to keep original table column types
+            sample_on_primary_table=False,  # go through normal sampling process with single table
+        )
+        sampled_table_name, _ = await self._get_or_cache_table(
```
YS-L, line 189: If the second element in the return value is False, it means that the table is not managed by `QueryCacheManagerService` and the caller is responsible for cleaning up the table. https://github.com/featurebyte/featurebyte/blob/483dc8ea2ff62e3fe34a0099382cccd0952cfef5/featurebyte/service/preview.py#L121-L131

I think the the flag needs to be handled.

featurebyte/query_graph/sql/interpreter/preview.py
```diff
@@ -448,6 +387,8 @@ def construct_sample_sql(
             Column to apply date range filtering on
         total_num_rows: int
             Total number of rows before sampling
+        sample_on_primary_table: bool
+            Whether to sample on primary table
```
YS-L, line 391: Is it more correct for this comment to be "Whether the graph is already sampled on primary table"? Since `_construct_sample_sql` will no longer perform sampling on primary table, right?

featurebyte/service/preview.py
```diff
@@ -299,6 +353,7 @@ async def sample(
             to_timestamp=sample.to_timestamp,
             timestamp_column=sample.timestamp_column,
             total_num_rows=total_num_rows,
+            sample_on_primary_table=True,
```
YS-L, line 356: Is it intended to always set this to True, vs having this as a flag like other methods?

tests/unit/query_graph/interpreter/test_preview.py
```diff
@@ -526,10 +545,20 @@ def test_describe__with_primary_table_sampling_on_graph_containing_filter(
         input_nodes=[node_join, node_eq],
     )
 
+    # identify primary table node & update the table details
```
YS-L, line 548: Would it be useful to make this a helper function in tests?

featurebyte/service/preview.py
gee-senbong: Got it. Do we need to handle it here too?
https://github.com/featurebyte/featurebyte-app/blob/c615af1c3e5e76e8d8adf04b1dd0d28e34c2780f/service/preview.py#L85

featurebyte/service/preview.py
gee-senbong: Original intention is to keep it the same with existing behaviour. If we make it an optional parameter, the parameters will be default to `True`, it will be inconsistent with other existing methods where `sample_on_primary_table` default to `False`. Wdyt? Should we make it an optional parameters with inconsistent default value?

featurebyte/service/preview.py
YS-L: Yes, I think that's an existing issue that should be fixed. Good catch!

featurebyte/service/preview.py
YS-L: Is there a reason the parameter has to default to `True` instead of `False`?

gee-senbong: `True` is the existing behaviour. Or we need to update all existing calls of the `sample` method. This is to prevent the case developers call it without providing the right parameter.

featurebyte/service/preview.py
```diff
@@ -152,6 +167,55 @@ async def _get_or_cache_dataframe(
...
+        sampled_table_name, is_table_cached = await self._get_or_cache_table(
+            session=session,
+            params=sample,
+            table_expr=sample_sql_tree,
+        )
+        output = sample.graph
+        for node in output.nodes:
+            if node.name == primary_table_input_node.name:
+                assert isinstance(node, InputNode)
+                node.parameters.table_details = TableDetails(table_name=sampled_table_name)
```
YS-L, line 211: This is mutating the graph passed in, right? Just checking this is the intended behaviour, as the resulting graph will not be valid for other purposes other than being used by this service. I'm mainly concerned about the usage in feature ideation.

featurebyte/service/preview.py
gee-senbong: You're right. I think i make a wrong assumption here. This is a very good catch! 🙏 🙏 🙏

YS-L (approved):

PR #2927, opened 2025-04-13
[chore] Fix feature type detection for non-embedding arrays

kchua78:
## Description

Fix feature type detection for non-embedding arrays and dict dtypes.

- Features of ARRAY type should not be detected as EMBEDDING feature since they do not have fixed lengths or contain only float values.
- Features of DICT and FLATDICT type (including deprecated MAP, STRUCT) should not be detected as  DICT feature since they may not be flat and may not have string keys or numeric values

## Type of Change

## Checklist

### DIFF
--- a/featurebyte/enum.py
+++ b/featurebyte/enum.py
@@ -156,6 +156,10 @@ class DBVarType(StrEnum):
     TIMEDELTA = "TIMEDELTA", "Time delta column"
     EMBEDDING = "EMBEDDING", "Embedding column"
     FLAT_DICT = "FLAT_DICT", "Flat dictionary column"
+    OBJECT = (
+        "OBJECT",
+        "Mixed-type column",
+    )  # used by aggregate features created from groupby with category parameter
 
     # specialized composite type
     TIMESTAMP_TZ_TUPLE = "TIMESTAMP_TZ_TUPLE", "Tuple of (timestamp, timezone offset)"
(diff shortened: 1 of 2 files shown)


featurebyte/enum.py
```diff
@@ -156,6 +156,10 @@ class DBVarType(StrEnum):
     TIMEDELTA = "TIMEDELTA", "Time delta column"
     EMBEDDING = "EMBEDDING", "Embedding column"
     FLAT_DICT = "FLAT_DICT", "Flat dictionary column"
+    OBJECT = (
+        "OBJECT",
+        "Mixed-type column",
+    )  # used by aggregate features created from groupby with category parameter
```
gee-senbong, line 162: Previously, @YS-L & have a discussion & we think that `OBJECT`, `DICT` and `MAP` have overlapping responsibility. Would be good to keep only `DICT` & deprecate the rest (long term plan). If we move it to here, does it mean we want to keep this `DBVarType`?

gee-senbong (approved):

featurebyte/enum.py
kchua78: We need something to differentiate between dict features produced from aggregation vs generic DICT from input tables. This is currently the type for the former and is required for detecting the DICT type that is supported for modeling. As discussed we will need to deprecate MAP, STRUCT and rename / refactor this separately

PR #3025, opened 2025-06-29
[DEV-4092] Add datetime partition column and schema to event and time series tables

kchua78:
## Description

Add datetime partition column and schema to event and time series tables

## Type of Change

## Checklist

### DIFF
--- a/featurebyte/query_graph/model/column_info.py
+++ b/featurebyte/query_graph/model/column_info.py
@@ -24,6 +24,17 @@ class ColumnSpecWithDescription(ColumnSpec):
     description: Optional[str] = Field(default=None)
 
 
+class ColumnSpecDetailed(ColumnSpecWithDescription):
+    """
+    ColumnInfo for storing column information with details
+    """
+
+    is_partition_key: bool = Field(
+        default=False,
+        description="Indicates if the column is a partition key",
+    )
+
+
 class ColumnInfo(ColumnSpecWithDescription):
     """
     ColumnInfo for storing column information
--- a/.changelog/DEV-4092.yaml
+++ b/.changelog/DEV-4092.yaml
@@ -1,0 +1,34 @@
+# This template file is used to generate changelog entries on release
+# Check the generated entry in your PR with the task command
+
+# To view the generated changelog, run the following command:
+# task changelog-pr
+
+# --- TEMPLATE --- #
+# One of 'breaking', 'deprecation', 'enhancement', 'bug_fix'
+change_type: enhancement
+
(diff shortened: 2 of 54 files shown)


featurebyte/query_graph/model/column_info.py
```diff
@@ -24,6 +24,17 @@ class ColumnSpecWithDescription(ColumnSpec):
     description: Optional[str] = Field(default=None)
 
 
+class ColumnSpecDetailed(ColumnSpecWithDescription):
```
YS-L, line 27: Is there a place where we need `ColumnSpecWithDescription` but not `ColumnSpecDetailed`? Wondering if it would simpler to keep just `ColumnSpecDetailed`.

kchua78: `ColumnInfo` is a subclass of `ColumnSpecWithDescription`, while we want to exclude `is_partition_key` from `ColumnInfo` to avoid changes to definition hashing.

Would it make sense if I remove `ColumnSpecWithDescription` and move `description` to both `ColumnInfo` and `ColumnSpecDetailed`?

featurebyte/query_graph/model/column_info.py
```diff
@@ -24,6 +24,17 @@ class ColumnSpecWithDescription(ColumnSpec):
...
+    ColumnInfo for storing column information with details
+    """
+
+    is_partition_key: bool = Field(
+        default=False,
+        description="Indicates if the column is a partition key",
+    )
+
+
 class ColumnInfo(ColumnSpecWithDescription):
```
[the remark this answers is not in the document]
kchua78: Would that affect definition hashing?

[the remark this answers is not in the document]
kchua78: Looks like `description` is already in `ColumnSpecWithDescription` so I can change it to inherit from `ColumnSpec` instead

featurebyte/query_graph/model/column_info.py
YS-L: `ColumnInfo` / `ColumnSpecWithDescription` are not part of the query graph. Only `ColumnSpec` is: https://github.com/featurebyte/featurebyte/blob/7a2d3ff57578e90fb394f0b21914de2a88cb3535/featurebyte/query_graph/node/input.py#L50

So none of these will affect graph hashing. Just a note: `is_partition_key` isn’t accessible to the query graph in the current implementation, in case that was your intention. I think that is fine. We can add new parameters to `BaseInputNodeParameters` with this information, which is probably the proper way to handle it.

PR #3084, opened 2025-08-05
[chore] Include datetime partition details in table get response

kchua78:
## Description

- Use different field `is_partition_key_candidate` to flag partition columns identified by database
- Populate datetime partition fields in get response of event and time series tables

## Type of Change

## Checklist


featurebyte/schema/table.py
```diff
@@ -197,3 +217,38 @@ class ColumnSemanticUpdate(FeatureByteBaseModel):
...
+        Any
+        """
+        columns_info = values.get("columns_info", [])
+        for column_info in columns_info:
+            partition_metadata = column_info.get("partition_metadata")
+            if partition_metadata is not None and partition_metadata.get("is_partition_key", False):
+                dtype_metadata = column_info.get("dtype_metadata")
+                if dtype_metadata is not None:
+                    values["datetime_partition_column"] = column_info["name"]
+                    values["datetime_partition_schema"] = dtype_metadata.get("timestamp_schema")
```
YS-L, line 252: Does it mean both of these fields will only be available when retrieving the model through the API route, but not at service level? Would it be better to include these fields at the table model level so that this information is more consistently available?

featurebyte/schema/table.py
```diff
@@ -69,6 +69,24 @@ def _special_column_validator(column_name: str, info: ValidationInfo) -> str:
...
+            Name of the field to set
+        value: Any
+            Value to set for the field
+        """
+        attr = getattr(self.__class__, name, None)
+        if isinstance(attr, property):
+            attr.__set__(self, value)
+        else:
+            self.__dict__[name] = value
+            self.__pydantic_fields_set__.add(name)
```
gee-senbong, line 88: Do we have test to cover this scenario? In case pydantic's underlying implementation changes, at least we have test to capture it.

featurebyte/schema/table.py
```diff
@@ -78,15 +96,17 @@ def _update_columns_info(self) -> "TableCreate":
         -------
         TableCreate
         """
-        if not self.datetime_partition_column:
-            return self
-
```
YS-L, line 83: If `datetime_partition_column` and `datetime_partition_schema` are available at the model level, I'm thinking we should probably update `construct_data_model_validator` to handle setting up these metadata in the columns info, instead of handling it here.

https://github.com/featurebyte/featurebyte/blob/526bd664e787cd5832c2a511788bb1dbe62e970b/featurebyte/models/event_table.py#L74-L82

We could have:

```python
 expected_column_field_name_type_pairs=[ 
     ("event_timestamp_column", DBVarType.supported_datetime_types()), 
     ("record_creation_timestamp_column", DBVarType.supported_timestamp_types()), 
     ("event_id_column", DBVarType.supported_id_types()), 
     ("event_timestamp_timezone_offset_column", {DBVarType.VARCHAR}), 
 ], 
 column_to_timestamp_schema_pairs=[ 
     ColumnToTimestampSchema("event_timestamp_column", "event_timestamp_schema"), 
 ], 
 datetime_partition_schema_pair=ColumnToTimestampSchema("datetime_partition_column", "datetime_partition_schema"),   # new, to implement
```

YS-L (commented):
> Use different field is_partition_key_candidate to flag partition columns identified by database

This part looks good to me.

> Populate datetime partition fields in get response of event and time series tables

I think the current approach works, but it’s starting to feel a bit hacky.

gee-senbong (approved):

featurebyte/schema/table.py
kchua78: Indeed it would be a lot simpler to have the fields in the model instead of doing these clean ups and conversion from column info. I will take that approach instead

featurebyte/schema/table.py
kchua78: There is a unit test for this class but we will not need this anymore once we store the datetime partition info in the model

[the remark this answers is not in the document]
kchua78: Will remove this as it’s too specific to pydantic implementation and easy to break

featurebyte/schema/table.py
YS-L: The latest change still updates the column info in this validator instead of in `construct_data_model_validator`. But I think it's already cleaner than before 👍

YS-L (approved):

YS-L (approved):

PR #3087, opened 2025-08-07
[chore] Support using empty existing table for external feature table

kchua78:
## Description

Support using empty existing table for external batch feature table.
If it's a DataBricks table it will be overwritten with update schema without replacement to preserve table location and permissions.

## Type of Change

## Checklist


featurebyte/service/batch_external_feature_table.py
```diff
@@ -251,6 +229,158 @@ async def _create_output_table(
...
+                        table=quoted_identifier("S"),
+                    ),
+                ),
+                expressions=[
+                    expressions.When(
+                        matched=True, then=expressions.Update(expressions=[expressions.Star()])
+                    ),
+                    expressions.When(
+                        matched=False, then=expressions.Insert(this=expressions.Star())
+                    ),
```
YS-L, line 356: Since the table is empty, this is the only branch that will be taken right?

YS-L (commented):
Is the new empty existing table flow covered by an integration test?

featurebyte/service/batch_external_feature_table.py
kchua78: Yes will it work without other branches?

kchua78:
> Is the new empty existing table flow covered by an integration test?

No, only unit. Will add to the current one

featurebyte/service/batch_external_feature_table.py
kchua78: It has to be empty since this logic will only be invoked if there are no columns in the table. And the intention is to overwrite any existing content

YS-L: I see, then I think we can just keep the insert branch to simpilfy.

featurebyte/service/batch_external_feature_table.py
```diff
@@ -135,12 +136,22 @@ async def get_batch_feature_table_task_payload(
                 schema_name=table_expr.db,
                 table_name=table_expr.name,
             )
-            available_column_details = {specs.name: specs for specs in column_specs}
 
-            # if table exists but has no columns, it is considered empty.
+            table_shape = await self.feature_store_warehouse_service.table_shape(
```
kchua78, line 140: Changed the empty table check logic to check for no rows instead of no columns as some databases do not support creation of a table without column

PR #3116, opened 2025-09-04
[chore] Use mongo for feature store cache storage

kchua78:
## Description

Use mongo for feature store cache storage
- Can be shared across replicas
- Can be managed (cleared) centrally

## Type of Change

## Checklist

### DIFF
--- a/featurebyte/models/feature_store_cache.py
+++ b/featurebyte/models/feature_store_cache.py
@@ -69,6 +69,7 @@ class FeatureStoreCacheModel(FeatureByteBaseDocumentModel):
             pymongo.operations.IndexModel("schema_name"),
             pymongo.operations.IndexModel("table_name"),
             pymongo.operations.IndexModel("key"),
+            pymongo.operations.IndexModel("value.function_name"),
             pymongo.operations.IndexModel("created_at", expireAfterSeconds=3600),
         ]
         auditable = False


featurebyte/models/feature_store_cache.py
```diff
@@ -0,0 +1,39 @@
...
+    class Settings(FeatureByteBaseDocumentModel.Settings):
+        """
+        Collection settings for FeatureStoreCacheModel
+        """
+
+        collection_name = "feature_store_cache"
+        unique_constraints = []
+        indexes = [
+            pymongo.operations.IndexModel("feature_store_id"),
+            pymongo.operations.IndexModel("key"),
```
YS-L, line 36: Should this be `keys`?

featurebyte/service/feature_store_warehouse.py
```diff
@@ -54,78 +55,48 @@
...
+        feature_store: FeatureStoreModel,
+        *args: Any,
+        refresh: bool = True,
+        **kwargs: Any,
+    ) -> T:
+        credentials = await self.session_manager_service.credential_service.get_credentials(
+            user_id=self.session_manager_service.user.id, feature_store=feature_store
+        )
+        keys = hashkey(func.__name__, credentials.id if credentials else None, *args, **kwargs)
+        str_keys = [str(key) for key in keys]
```
YS-L, line 77: Is there a reason to keep a list of keys instead of hashing everything into a single string?

featurebyte/service/feature_store_warehouse.py
```diff
@@ -54,78 +55,48 @@
...
+        query_filter = {"feature_store_id": feature_store.id, "keys": str_keys}
+        cached_results: List[
+            FeatureStoreCacheModel
+        ] = await self.feature_store_cache_service.list_documents(query_filter=query_filter)
```
YS-L, line 81: If `refresh=True` this mongo query would be redundant.

featurebyte/models/feature_store_cache.py
```diff
@@ -0,0 +1,48 @@
...
+import pymongo
+from pydantic import Field
+
+from featurebyte.models.base import FeatureByteBaseDocumentModel, PydanticObjectId
+from featurebyte.query_graph.model.column_info import ColumnSpecWithDescription
+from featurebyte.query_graph.model.table import TableDetails, TableSpec
+
+FeatureStoreCachedValue = Union[
+    List[str], List[TableSpec], List[ColumnSpecWithDescription], TableDetails, dict[str, str]
+]
```
YS-L, line 16: This seems a bit fragile. What happens when the return types of functions change over time, or when we choose to cache new functions?

I'm thinking adding a bit more structure / type safety might help: make this a union over concrete result type variants with `function_name` as the discriminator, something like:

```python
  class DatabaseListCache(BaseModel):
      function_name: Literal["list_databases"] = "list_databases"
      value: List[str]

  class SchemaListCache(BaseModel):
      function_name: Literal["list_schemas"] = "list_schemas"
      value: List[str]

  class TableListCache(BaseModel):
      function_name: Literal["list_tables"] = "list_tables"
      value: List[TableSpec]

  class ColumnListCache(BaseModel):
      function_name: Literal["list_columns"] = "list_columns"
      value: List[ColumnSpecWithDescription]

  class TableDetailsCache(BaseModel):
      function_name: Literal["get_table_details"] = "get_table_details"
      value: TableDetails

  CacheValue = Annotated[
      Union[
          DatabaseListCache,
          SchemaListCache,
          TableListCache,
          ColumnListCache,
          TableDetailsCache
      ],
      Field(discriminator="function_name")
  ]

```

Then we have some control over backward compatibility if changes occur over time, and it's clear which function is associated with which return type.

Btw, why do we need `dict[str, str]`?

featurebyte/service/feature_store_warehouse.py
```diff
@@ -54,78 +55,51 @@
...
-        @functools.wraps(func)
-        async def wrapper(
-            self: FeatureStoreWarehouseService,
-            feature_store: FeatureStoreModel,
-            *args: Any,
-            refresh: bool = True,
-            **kwargs: Any,
-        ) -> T:
-            if refresh:
-                cache_dict[feature_store.id].clear()
```
YS-L, line 113: Previously, `refresh=True` would clear all keys associated with the cache, but now it only clears one specific key. Is that an intended behaviour change?

kchua78: yes it's an intended change. Refresh is used in operations to ensure the latest info is fetched for a specific request. It seems to harsh to clear the cache for the entire feature store.

kchua78: There's a separate route to clear the entire store cache for use cases where that is the intention

featurebyte/models/feature_store_cache.py
kchua78: It's for extended functionality not available in this repo. I will remove it and extend the model elsewhere

YS-L (approved):

PR #3340, opened 2026-04-27
Dedup concurrent feature compute on the same observation table

Gxav:
@YS-L 
Move the per-observation-table lock in
FeatureTableCacheService.create_or_update_feature_table_cache to wrap the entire cache-check + compute + insert sequence, instead of just the insert. Without this, two concurrent requests for the same (observation_table, features) pair can both miss the cache, both decide to compute, and both run the expensive feature computation in parallel — the previous insert-only lock protected against concurrent writes but did nothing to dedup the duplicated work.

Observed in production: two parallel model training tasks computing identical features against the same observation table did ~85 minutes of duplicated execution and ~34 GB of duplicated scan, simply because they raced past the cache check before either could populate the cache.

The inner lock in _materialize_and_update_cache is now redundant and removed.

Adds a regression test that runs two concurrent
create_or_update_feature_table_cache calls via asyncio.gather and asserts that the underlying feature compute fires exactly once.

## Type of Change

## Checklist

### DIFF
--- a/featurebyte/service/feature_table_cache.py
+++ b/featurebyte/service/feature_table_cache.py
@@ -690,17 +690,26 @@ class FeatureTableCacheService:
                     CACHE_INSERTION_START_PROGRESS_PERCENTAGE, "Inserting features into cache"
                 )
 
-            # No lock needed here — the caller (create_or_update_feature_table_cache)
-            # already holds the per-observation-table lock that wraps cache check +
-            # compute + insert.
-            await self._insert_features_into_cache(
-                db_session=db_session,
-                observation_table=observation_table,
-                non_cached_nodes=non_cached_nodes,
-                graph=graph,
-                intermediate_table_name=intermediate_table_name,
-                features_computation_result=features_computation_result,
-            )
+            # Defense in depth around the cache table mutation (create / alter / merge).
+            # The outer lock in create_or_update_feature_table_cache normally serializes
+            # this, but its TTL is finite (LONG_RUNNING_EXECUTE_QUERY_TIMEOUT_SECONDS) —
+            # if it expires mid-flight a concurrent caller could re-enter the insert
+            # path. We use a key distinct from the outer lock so re-acquiring here
+            # doesn't deadlock against this same process's own outer lock (Redis locks
+            # are not reentrant).
+            async with acquire_lock(
+                self.redis,
+                f"feature_table_cache_insert:obs:{observation_table.id}",
+                timeout=LONG_RUNNING_EXECUTE_QUERY_TIMEOUT_SECONDS,
+            ):
+                await self._insert_features_into_cache(
+                    db_session=db_session,
+                    observation_table=observation_table,
+                    non_cached_nodes=non_cached_nodes,
+                    graph=graph,
+                    intermediate_table_name=intermediate_table_name,
+                    features_computation_result=features_computation_result,
+                )
 
             # Cache insertion completed
             if progress_callback:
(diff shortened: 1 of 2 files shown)


featurebyte/service/feature_table_cache.py
```diff
@@ -690,21 +690,17 @@ async def _materialize_and_update_cache(  # pylint: disable=too-many-arguments
...
-                    observation_table=observation_table,
-                    non_cached_nodes=non_cached_nodes,
-                    graph=graph,
-                    intermediate_table_name=intermediate_table_name,
-                    features_computation_result=features_computation_result,
-                )
+            # No lock needed here — the caller (create_or_update_feature_table_cache)
+            # already holds the per-observation-table lock that wraps cache check +
+            # compute + insert.
+            await self._insert_features_into_cache(
```
YS-L: If the caller's lock expires (it has a timeout of `LONG_RUNNING_EXECUTE_QUERY_TIMEOUT_SECONDS`), this becomes unprotected. I think we should keep this lock here as a defensive layer.

featurebyte/service/feature_table_cache.py
```diff
@@ -810,43 +803,65 @@ async def create_or_update_feature_table_cache(
...
+        # (observation_table, features) pair can both miss the cache, both decide to
+        # compute, and both run the expensive feature computation in parallel — even
+        # though the existing inner insert-only lock protects against concurrent writes,
+        # it doesn't dedup the compute. Holding the lock from the cache check through
+        # insert means a second request finds the cache populated and exits cheaply.
+        async with acquire_lock(
+            self.redis,
+            f"feature_table_cache_update:obs:{observation_table.id}",
+            timeout=LONG_RUNNING_EXECUTE_QUERY_TIMEOUT_SECONDS,
+        ):
```
YS-L: This disables parallel queries even for completely unrelated features on the same observation table. Is this the desired tradeoff?

Gxav: you are right. it is not the desired tradeoff. I will address this

featurebyte/service/feature_table_cache.py
```diff
@@ -810,43 +812,73 @@ async def create_or_update_feature_table_cache(
...
-        if progress_callback:
-            await progress_callback(
-                FEATURE_TABLE_CACHE_CHECK_PROGRESS_PERCENTAGE,
-                "Feature table cache status check completed",
+        # Serialize the cache-check + compute + insert sequence per observation table.
+        # Without this lock at the outer level, two concurrent requests for the same
+        # (observation_table, features) pair can both miss the cache, both decide to
+        # compute, and both run the expensive feature computation in parallel.
+        # Holding the lock from the cache check through insert means a second request
+        # finds the cache populated and exits cheaply.
```
YS-L: Is this comment accurate?

For non-overlapping feature requests, the second request will be blocked. Where does "exists cheaply" come in and what does it mean exactly?

featurebyte/service/feature_table_cache.py
YS-L: It seems it's the accepted tradeoff in the final implementation? Could you confirm it's intended?

YS-L (approved):

featurebyte/service/feature_table_cache.py
Gxav: Actually. It should not accept this tradeoff. I will update my PR. Sorry for this.

Gxav:
@YS-L Good call on the parallelism concern. I pushed a fix.

1. The outer lock is now keyed on (observation_table.id, sha256-fingerprint of sorted unique definition hashes) instead of just observation_table.id. So two parallel requests for the same feature list on the same observation table still share a lock and dedup (the original intent / the ~85min ~34GB production case), but unrelated feature sets on the same observation table acquire different locks and run in parallel.

2. I added test_create_or_update_feature_table_cache__different_feature_sets_run_in_parallel that runs two concurrent requests with disjoint single-node subsets and asserts max_concurrent == 2. Verified it fails when the lock is reverted to per-observation-only, so it actually catches the regression

